PR #7, opened 2021-06-14
LTG-35: basic unit test for enter phone number route

dbes-gds:
## What

Basic unit test for enter-phone-number route.

## Why

Seed unit test creation for subsequent routes.


tests/unit/routes/enter-phone-number.test.ts
```diff
@@ -0,0 +1,15 @@
+const assert = require("assert");
+
+import request from "supertest";
+import { createApp } from "../../../src/app";
+
+describe("Unit testing the /enter-phone-number GET route", () => {
+  it("should return OK status", () => {
+    const app = createApp();
```
sanisc, line 8: this uses promises as its syntax it might be worthwhile changing it to opposite e.g 

```
request(createApp())
      .get("/enter-phone-number")
      .expect(200)
.end(function(err, res) {
        if (err) return done(err);
        return done();
      });
```

sanisc (changes requested):

tests/unit/routes/enter-phone-number.test.ts
dbes-gds: updated

sanisc (approved):

PR #14, opened 2021-06-18
LTG-35: add account created successfully page

dbes-gds:
## What

Add 'account created successfully' page.
When 2FA not required navigate to this page after 'enter password', otherwise navigate to 'enter phone number'.
In order to test the flow include 'no2fa' in the email to simulate a client not requiring 2FA.

## Why

- Given that the user has successfully created their password AND the service that the user requires only Email and Password for to access their service, When they select “Continue”, Then an account must be created for the user AND they should be directed to “You have successfully created your GOV.UK account” UI.

### DIFF
--- a/src/controllers/register-create-password-controller.ts
+++ b/src/controllers/register-create-password-controller.ts
@@ -73,9 +73,12 @@ export function createAccountPost(
           req.body["password"]
         )
       ) {
-        return res.redirect(PATH_NAMES.CREATE_ACCOUNT_ENTER_PHONE_NUMBER);
+        if (req.session.user.email.includes("no2fa")) {
+          return res.redirect(PATH_NAMES.CREATE_ACCOUNT_SUCCESSFUL);
+        } else {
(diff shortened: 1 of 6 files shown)


src/controllers/register-create-password-controller.ts
```diff
@@ -73,9 +73,12 @@ export function createAccountPost(
           req.body["password"]
         )
       ) {
-        return res.redirect(PATH_NAMES.CREATE_ACCOUNT_ENTER_PHONE_NUMBER);
+        if (req.session.user.email.includes("no2fa")) {
```
sanisc, line 76: is this not a property that comes back from the web request? This on the email in the session

src/routes.ts
```diff
@@ -76,6 +77,17 @@ router.get(
   enterPhoneNumberGet
 );
 
+router.get(
```
sanisc, line 80: its probably worth adding a basic test to test the get request

sanisc (approved):

PR #18, opened 2021-06-21
BAU: add script to launch the app locally with a session id

dbes-gds:
## What 

Add script to launch the app locally with a session id

## Why

If running a local node app against the remote API a session id is required from the API in order to run the flow.  This script retrieves the session id and passes it to the local app.

### DIFF
--- a/session-start-local.sh
+++ b/session-start-local.sh
@@ -1,0 +1,12 @@
+#!/usr/bin/env bash
+set -eu
+
+set -o allexport
+source .env
+set +o allexport
+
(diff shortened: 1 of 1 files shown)


JHjava (approved):

session-start-local.sh
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env bash
+set -eu
+
+set -o allexport
+source .env
```
JHjava, line 5: Is there meant to be a `.env` file?

sanisc: you add the .env file in yourself locally but its never checked in

JHjava: Ah yep. I see that now

PR #52, opened 2021-06-24
LTG-371: tweak 'check your email' to look like prototype

dbes-gds:
## What?

Tweak 'check your email' to look like prototype.
Make sure all text and fonts are correct.
Include the email address entered on the previous page.

## Why?

Page must conform to the design.

### DIFF
--- a/src/components/common/layout/base.njk
+++ b/src/components/common/layout/base.njk
@@ -18,6 +18,21 @@
   <![endif]-->
 {% endblock %}
 
+{% block pageTitle %}
+    {% if error or errors %}
+        {{ 'general.errorTitlePrefix' | translate }}
+        -
+    {% endif %}
+    {% if pageTitleName %}
+        {{ pageTitleName }}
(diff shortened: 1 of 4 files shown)


alan-gds (commented):
Basically fine, but before we merge this, I want to get to the bottom of what's going on with that `#SOMETHING` placeholder I wrote in the nunjucks template.  The replacement seems to be necessary, yet the placeholder is compeltely arbitrary and is not actually referenced anywhere else.  This was also the case for the template that we copied the pattern from (register-create-password/index.njk).  The `#PRIVACY_NOTICE_LINK` placeholder there is never actually referenced anywhere - I would expect it to be in the translations file.

Until we understand the significance of these placeholders and give the new one a better name, we shouldn't merge.

alan-gds (approved):
Sandeep's changes resolve my concerns about the `replace` expression.

alan-gds (approved):

PR #59, opened 2021-06-28
Connect verify-code backend API call to UI

sanisc:
## What?

Hook up backend API call for verify code to front end logic

## Why?

This is needed so the user can verify the email code they are sent and continue on with their journey.


dbes-gds (approved):

src/components/verify-email/verify-email-routes.ts
```diff
@@ -1,25 +1,22 @@
...
-  registerVerifyEmailGet
+  verifyEmailGet
 );
 
 router.post(
   PATH_NAMES.CHECK_YOUR_EMAIL,
   basicMiddlewarePipeline,
   validateVerifyEmailRequest(),
-  registerVerifyEmailPost
+  verifyEmailPost()
```
dbes-gds, line 19: brackets needed?

sanisc: yes , brackets are needed as its calling a higher order function

dbes-gds (approved):

PR #63, opened 2021-06-29
BAU: update session-start-local script

dbes-gds:
## What?

Update session-start-local script.

## Why?

Due to changed /authorize parameters.


alan-gds (approved):
I don't fully follow the details of this change, but it looks sensible on the face of it and it does fix the script - for me it currently does nothing in the main branch, and successfully starts the session in this PR branch.

PR #64, opened 2021-06-29
LTG-405: Connect 'enter password' page to backend

alan-gds:
## What?

Connect the 'enter password' page to the backend API.

## Why?

So that it can interoperate with the backend.


sanisc (dismissed):
couple of linting issues but other than that looks good.

sanisc:
also i don't think the page title is set in the html page so its worth adding that in

dbes-gds (approved):

PR #76, opened 2021-07-02
LTG-375: Update content and validation for 'Enter Phone Number'

dbes-gds:
## What?

LTG-375: Update content and validation for 'Enter Phone Number'.

## Why?

Align with the latest content.
Add phone number validation rules to support UK fixed and mobile numbers using 'libphonenumber-js'.


src/utils/phone-number.ts
```diff
@@ -0,0 +1,14 @@
...
+
+export function containsUKMobileNumber(value: string): boolean {
+  return isValidPhoneNumber(value, "GB");
+}
+
+export function containsNumbersOrSpacesOnly(value: string): boolean {
+  return value ? /^[\d\s]+$/.test(value) : false;
+}
+
+export function lengthInRangeWithoutSpaces(value: string, min: number, max: number): boolean {
```
sanisc, line 11: can we add some unit tests around this to make sure it works as expected

sanisc (changes requested):

src/utils/phone-number.ts
dbes-gds: added

sanisc (approved):

PR #87, opened 2021-07-06
GOVSI-429: Basic page for security code expired

dbes-gds:
## What?

Basic page for security code expired
Redirects to the page when the verifyPhoneNumber API returns number of tries exceeded.

Changes may be required when the content is confirmed.

## Why?

User gets 5 attempts to enter the security code correctly.


src/components/security-code-expired/security-code-expired-controller.ts
```diff
@@ -0,0 +1,9 @@
+import { Request, Response } from "express";
+
+export function securityCodeExpiredGet(req: Request, res: Response): void {
+  res.render("security-code-expired/index.njk");
+}
+
+export function securityCodeExpiredPost(req: Request, res: Response): void {
```
sanisc, line 7: do we need this post method?

src/components/security-code-expired/security-code-expired-routes.ts
```diff
@@ -0,0 +1,23 @@
...
+import { validateSessionMiddleware } from "../../middleware/session-middleware";
+
+const router = express.Router();
+
+router.get(
+  PATH_NAMES.SECURITY_CODE_EXPIRED,
+  validateSessionMiddleware,
+  securityCodeExpiredGet
+);
+router.post(
```
sanisc, line 17: same as above

src/components/security-code-expired/security-code-expired-controller.ts
dbes-gds: removed pending clarification on content + behaviour

src/components/security-code-expired/security-code-expired-routes.ts
dbes-gds: removed pending clarification on content + behaviour

sanisc (approved):

PR #103, opened 2021-07-13
GOVSI-104: add cookie policy page and save preferences

dbes-gds:
## What?

Update the cookie policy page with the latest content.
Add required UI components.
Save updated preferences when the user clicks the save button.

## Why?

As a user 
I want to know how the cookies will be used on the authentication service 
So that I can make an informed decision as to whether I want the service to use cookies

### DIFF
--- a/src/assets/javascript/cookies.js
+++ b/src/assets/javascript/cookies.js
@@ -56,14 +56,14 @@ window.govSigin = window.govSigin ?? {};
     });
   }
 
-  function setCookie(n, v) {
+  function setCookie(name, value) {
     var currentDate = new Date();
     var expiryDate = new Date(
       currentDate.setMonth(currentDate.getMonth() + 12),
     );
     document.cookie =
-      n + "=" +
-      JSON.stringify(v) +
+      name + "=" +
+      JSON.stringify(value) +
       "; expires=" +
       expiryDate +
       "; path=/; Secure";
@@ -82,7 +82,7 @@ window.govSigin = window.govSigin ?? {};
   }
 
   function cookiesPageInit() {
-    var cookie = getCookie(COOKIES_PREFERENCES_SET);
+    var cookie = getCookieValue(COOKIES_PREFERENCES_SET);
     var analyticsValue = false;
 
     if (!cookie) {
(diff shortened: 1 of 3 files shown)


src/assets/javascript/cookies.js
```diff
@@ -13,10 +13,12 @@ window.govSigin = window.govSigin ?? {};
   var acceptCookies = document.querySelector("button[name=\"cookiesAccept\"]");
   var rejectCookies = document.querySelector("button[name=\"cookiesReject\"]");
   var cookiePreferencesExist =
-    document.cookie.indexOf(COOKIES_PREFERENCES_SET + '=') > -1;
+    document.cookie.indexOf(COOKIES_PREFERENCES_SET + "=") > -1;
 
   function init() {
-    if (cookiePreferencesExist || isOnCookiesPage()) {
+    if (isOnCookiesPage()) {
```
sanisc, line 19: i would split the method into two and keep the boolean check separate. e.g

```
 if (isOnCookiesPage()) {
      cookiesPageInit();
      return;
    } else if (cookiePreferencesExist) {
      return;
    }
```

src/assets/javascript/cookies.js
```diff
@@ -75,7 +77,50 @@ window.govSigin = window.govSigin ?? {};
   }
 
   function isOnCookiesPage() {
```
sanisc, line 79: further to the comment above a way to spilt it could be as follows:

```
function cookiesPageInit() {
    var cookie = getCookie(COOKIES_PREFERENCES_SET);
    var analyticsValue = false;

    if (!cookie) {
      setCookie(COOKIES_PREFERENCES_SET, { analytics: false });
    } else {
      analyticsValue = JSON.parse(cookie).analytics;
    }

    document.querySelector("#policy-cookies-accepted").checked = analyticsValue;
    document.querySelector("#policy-cookies-rejected").checked =
      !analyticsValue;

    document.querySelector("#save-cookie-settings").addEventListener(
      "click",
      function (event) {
        event.preventDefault();
        var selectedPreference = document.querySelector(
          '#radio-cookie-preferences input[type="radio"]:checked'
        ).value;
        setCookie(COOKIES_PREFERENCES_SET, {
          analytics: selectedPreference === "true",
        });
        showElement(document.querySelector("#save-success-banner"));
        window.scrollTo(0, 0);
      }.bind(this)
    );

    document.querySelector("#go-back-link").addEventListener(
      "click",
      function (event) {
        event.preventDefault();
        window.history.back();
      }.bind(this)
    );
  }

  function isOnCookiesPage() {
    return window.location.pathname !== "/cookies";
  }
```

src/assets/javascript/cookies.js
```diff
@@ -78,5 +81,47 @@ window.govSigin = window.govSigin ?? {};
...
+      "click",
+      function(event) {
+        event.preventDefault();
+        window.history.back();
+      }.bind(this),
+    );
+  }
+
+  function getCookie(n) {
+    var c = document.cookie
```
sanisc, line 116: can we change the variable name to what it is e.g 

`var cookies = document.cookie`

sanisc (changes requested):

sanisc (approved):

PR #126, opened 2021-07-22
GOVSI-153: create signed out page

dbes-gds:
## What?

Create signed out page.

## Why?

As a user
I want confirmation that I have signed out of my account 
So that I am confident my information is safe

### DIFF
--- a/src/components/signed-out/signed-out-controller.ts
+++ b/src/components/signed-out/signed-out-controller.ts
@@ -1,0 +1,5 @@
+import { Request, Response } from "express";
+
+export function signedOutGet(req: Request, res: Response): void {
+  res.render("signed-out/index.njk");
+}
--- a/src/components/signed-out/signed-out-controller.ts
+++ b/src/components/signed-out/signed-out-controller.ts
@@ -1,5 +1,12 @@
 import { Request, Response } from "express";
 
 export function signedOutGet(req: Request, res: Response): void {
+  if (req.cookies) {
+    for (const cookieName of Object.keys(req.cookies)) {
+      if (cookieName !== 'lng' && cookieName !== 'cookies_preferences_set' ) {
+        res.clearCookie(cookieName);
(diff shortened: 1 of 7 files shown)


src/components/signed-out/signed-out-routes.ts
```diff
@@ -0,0 +1,18 @@
...
+import {
+  signedOutGet,
+} from "./signed-out-controller";
+import { validateSessionMiddleware } from "../../middleware/session-middleware";
+
+const router = express.Router();
+
+router.get(
+  PATH_NAMES.SIGNED_OUT,
+  validateSessionMiddleware,
```
sanisc, line 13: this is not needed as they won't have a session at this point

src/components/signed-out/signed-out-controller.ts
```diff
@@ -0,0 +1,5 @@
+import { Request, Response } from "express";
+
+export function signedOutGet(req: Request, res: Response): void {
+  res.render("signed-out/index.njk");
```
sanisc, line 4: i would clear all cookies on the res if there are any

sanisc (changes requested):

src/components/signed-out/signed-out-controller.ts
dbes-gds: Which cookies? Guess we should leave cookie preferences?

src/components/signed-out/signed-out-routes.ts
dbes-gds: removed

src/components/signed-out/signed-out-controller.ts
dbes-gds: deleting all cookies except 'cookies_preferences_set' and 'lng' as these are user preference rather than session related.

sanisc (approved):

PR #135, opened 2021-07-30
Remove session id from app cookie

sanisc:
## What & Why?

Remove the session id from app cookie - this is given back in another cookie and a piece of middleware has been added to handle reading the session id from the new cookie.

Auth-code redirect url has been added so it happens in the user agent rather on the server side.


src/components/enter-phone-number/tests/enter-phone-number-controller.test.ts
```diff
@@ -41,9 +41,9 @@ describe("enter phone number controller", () => {
         updateProfile: sandbox.fake.returns(true),
       };
 
+      res.locals.sessionId = "123456-djjad";
       req.body.phoneNumber = "07738393990";
       req.session.user = {
-        sessionId: "dadll-33",
         email: "test@test.com",
         scope: "openid",
```
JHjava, line 48: Does this need to be deleted?

sanisc: good spot and yes it does

JHjava (approved):

PR #165, opened 2021-08-17
GOVSI-281: updated content for 'share-info' page

dbes-gds:
## What?

By default no option will be selected.
On clicking continue an error message will be displayed if no option is selected.

## Why?

Page content has been updated in the Figma and there is new default behaviour and an error message for when no option is selected by the user for sharing.

## Related PRs

#155 
#267
#265
#261

### DIFF
--- a/src/components/share-info/index.njk
+++ b/src/components/share-info/index.njk
@@ -52,18 +52,19 @@
   },
   items: [
     {
-      checked: "false",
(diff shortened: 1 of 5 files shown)


sanisc (changes requested):
can we add some tests for this

dbes-gds:
> can we add some tests for this

added integration tests

sanisc (approved):

PR #183, opened 2021-08-23
Front end changes for mandatory and optional terms and conds

cj1993:
## What?

Front end changes for mandatory and optional terms and conds


src/components/updated-terms-conds/updated-terms-conds-controller.ts
```diff
@@ -1,52 +1,81 @@
...
+    return async function (req: Request, res: Response) {
+        const clientInfo = await callClientInfo(req, res, service);
+        req.session.redirectUri = clientInfo.redirectUri;
+        req.session.state = clientInfo.state;
+
+        res.render("updated-terms-conds/index.njk");
+    };
+}
+
+export function updatedTermsCondsMandatoryGet(
```
sanisc, line 19: is there a reason for having 2 different routes for each page could it not be done under one? i would speak to suly about this as the page urls are usually defined in figma and eventually will be tracked by GA

cj1993: As discussed, Suly & Kim have confirmed that they want 3 separate URL paths for GA tracking.

src/components/updated-terms-conds/updated-terms-conds-controller.ts
```diff
@@ -1,52 +1,81 @@
...
-  };
+    return async function (req: Request, res: Response) {
+        const updatedTermsAndCondsValue = true;
+        const {email} = req.session.user;
+        const sessionId = res.locals.sessionId;
+        const redirectUri = req.session.redirectUri;
+        const state = req.session.state;
+        const acceptOrReject = req.body.acceptOrReject;
+
+        if (acceptOrReject === "reject") {
```
sanisc, line 58: you only want to add state is the session has something this would add an empty param?

cj1993: The state gets set on the session from the calls to the clientInfo lambda on any of the 3 get requests, so I don't think that field will be empty.

sanisc (approved):

PR #185, opened 2021-08-24
GOVSI-457: research 5 sign-in flow integration

dbes-gds:
## What?

Put the updated research 5 sign-in flow live.

'sign-in-or-create' becomes the new landing page.
Subsequent pages updated to follow the new flow.

GOVSI-457: create 'sign-in-or-create' page
GOVSI-458: create new version of 'enter-email with two modes for new and existing user
GOVSI-459: create new 'account-not-found' page
GOVSI-691: create new version of 'enter-password' when user has an account already

The new flow also supports both versions of the screens for the OPTIONAL and MANDATORY flow.  However, MANDATORY is the current main option and is used by default.  Screens for OPTIONAL are subject to further change and should not be tested as part of these tickets.

## Why?

User journey has been updated as a result of user research round 5.

## Related PRs

#173


src/components/sign-in-or-create/sign-in-or-create-controller.ts
```diff
@@ -1,12 +1,28 @@
...
-export function signInOrCreateGet(req: Request, res: Response): void {
-  res.render("sign-in-or-create/index.njk");
+export function signInOrCreateGet(
+  service: SignInOrCreateServiceInterface = signInOrCreateService(),
+): ExpressRouteFunc {
+  return async function(req: Request, res: Response) {
+    const sessionId = res.locals.sessionId;
+    const clientSessionId = res.locals.clientSessionId;
+    const clientInfo = await service.clientInfo(sessionId, clientSessionId);
+    if (clientInfo.service_type !== undefined || clientInfo.service_type === SERVICE_TYPE.MANDATORY || clientInfo.service_type === SERVICE_TYPE.OPTIONAL) {
```
sanisc, line 14: this could be potentially simplified to

```
if(clientInfo.serviceType){
.....
```

src/components/enter-password/enter-password-controller.ts
```diff
@@ -45,7 +51,7 @@ export function enterPasswordPost(
       "password",
       req.t("pages.enterPassword.password.validationError.incorrectPassword")
     );
-
+    //TODO this won't work properly
     renderBadRequest(res, req, TEMPLATE_NAME, error);
```
sanisc, line 55: has this been addressed? the todo

sanisc (changes requested):

src/components/enter-password/enter-password-controller.ts
dbes-gds: Fixed, validation works for both enter password pages.

PR #206, opened 2021-09-15
GOVSI-128: Add Google Tag Manager scripts

alan-gds:
## What?

Add Google Tag Manager scripts to the header in all pages, along with a hidden iframe at the top of the page body, via the base template.

## Why?

The Performance Team will be using Google Tag Manager for analytics.  I'm told the hidden iframe is part of the GTM magic.  See the JIRA ticket for more details.

### DIFF
--- a/src/assets/javascript/cookies.js
+++ b/src/assets/javascript/cookies.js
@@ -1,8 +1,6 @@
-window.govSigin = window.govSigin ?? {};
+"use strict";
 
-(function (w) {
-  "use strict";
-
+var cookies = function (trackingId) {
   var COOKIES_PREFERENCES_SET = "cookies_preferences_set";
 
   var cookiesAccepted = document.querySelector("#cookies-accepted");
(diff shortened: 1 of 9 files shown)


src/components/common/layout/base.njk
```diff
@@ -3,6 +3,20 @@
 {% from "govuk/components/phase-banner/macro.njk" import govukPhaseBanner %}
 
 {% block head %}
+    <script>
```
sanisc, line 6: this won't work as the CSP blocks any inline scripts from running for security reasons

src/assets/javascript/cookies.js
```diff
@@ -1,8 +1,6 @@
-window.govSigin = window.govSigin ?? {};
-
-(function (w) {
-  "use strict";
+"use strict";
 
+var cookies = function (trackingId) {
```
alan-gds, line 3: Indentation looks off here.

src/assets/javascript/cookies.js
```diff
@@ -55,80 +38,180 @@ window.govSigin = window.govSigin ?? {};
...
+      showElement(cookiesRejected);
+    }
   }
 
-  function isOnCookiesPage() {
-    return window.location.pathname === "/cookies";
+  function saveCookieSettings(event) {
+    event.preventDefault();
+
+    const hasConsented =
```
alan-gds, line 69: Rogue const in .js file

src/assets/javascript/cookies.js
```diff
@@ -55,80 +38,180 @@ window.govSigin = window.govSigin ?? {};
...
+      setCookie(gtagCookie, "", { days: -1 });
+    }
+
+    window.scrollTo(0, 0);
   }
 
   function cookiesPageInit() {
-    var cookie = getCookieValue(COOKIES_PREFERENCES_SET);
-    var analyticsValue = false;
+    const analyticsConsent = hasConsentForAnalytics();
```
alan-gds, line 97: Rogue const in .js file

alan-gds (commented):
Think this is generally okay, apart from some consts being present in a .js file.  I only checked what's in the diff, and there were already a few in there, so worth doing a search to see if there are others we have missed.

I've also flagged an indentation issue but that doesn't matter much.

src/assets/javascript/cookies.js
alan-gds: My mistake; I misinterpreted the diff.

alan-gds (commented):
LGTM

sanisc (approved):

PR #233, opened 2021-09-22
Add account locked scenario

sanisc:
## What?

Add account locked out page

## Why?

When a user exceeds the number of attempts to end a password they should be prompted with a error page.


src/app.constants.ts
```diff
@@ -43,6 +44,20 @@ export const HTTP_STATUS_CODES = {
...
+  SIGNUP_USER: "/signup",
+  SEND_NOTIFICATION: "/send-notification",
+  VERIFY_CODE: "/verify-code",
+  LOG_IN_USER: "/login",
+  UPDATE_PROFILE: "/update-profile",
+  MFA: "/mfa",
+  AUTH_CODE: "/auth-code",
+  CLIENT_INFO: "/client-info",
+  RESET_PASSWORD_REQUEST: "/reset-password-request",
+  RESET_PASSWORD: "/reset-password",
```
JHjava, line 58: Are these meant to be here?

sanisc: Yes its the endpoint for the API

JHjava (approved):

PR #269, opened 2021-10-04
Update content to match figma

sanisc:
## What?

Update content to match figma final designs.


sanisc:
ignore the welsh file its has been just kept up to date with en file

JHjava (approved):

PR #391, opened 2021-11-16
GOVSI-1099 - Validate against top 100k passwords

JHjava:
## What?

- Use the PwnedPasswordsTop100k as recommended by NCSC to validate against new passwords or reset passwords. If the password is in the list then return an error. https://www.ncsc.gov.uk/blog-post/passwords-passwords-everywhere
- Amend package.json to copy over the txt file to dist
- Amend nodemon.json so it picks up changes to the PwnedPasswordsTop100k.txt on the fly.
- Add a custom validator which calls a new function to validate the password and add unit tests to check logic.

## Why?

- To ensure that users use stronger passwords which are less likely to be compromised.


sanisc (changes requested):
in the docker container run - yarn pretty

test/unit/utils/password-validation.test.ts
```diff
@@ -0,0 +1,16 @@
...
+    isCommonPassword,
+} from "../../../src/utils/password-validation";
+
+describe("password-validation", () => {
+    describe("commonPassword", () => {
+    it("should return true when password is in common password list", () => {
+        expect(isCommonPassword("password")).to.equal(true);
+      });
+    it("should return false when password is not in common password list", () => {
+    expect(isCommonPassword("arthnfk174_g2")).to.equal(false);
```
sanisc, line 13: what happens if the password is null or ""?

JHjava: It wouldn't get that far due to the ordering of the `create-password-validation` we check for lengths etc before checking for common passwords

sanisc (approved):

PR #405, opened 2021-11-23
GOVSI-1142: read 'di-persistent-session-id' cookie and pass to api calls as header

dbes-gds:
## What?

Read 'di-persistent-session-id' cookie and pass to api calls as header.

- Cookie value is read by the middleware and saved into 'res.locals', then retrieved for each api call.

## Why?

Cookie value to be logged by each api.

## Related PRs

https://github.com/alphagov/di-authentication-api/pull/1087

### DIFF
--- a/src/components/common/client-info/types.ts
+++ b/src/components/common/client-info/types.ts
@@ -17,6 +17,7 @@ export interface ClientInfoServiceInterface {
   clientInfo: (
     sessionId: string,
     clientSessionId: string,
-    sourceIp: string
+    sourceIp: string,
+    persistentSessionId: string
   ) => Promise<ClientInfoResponse>;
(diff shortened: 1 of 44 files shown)


src/components/common/client-info/types.ts
```diff
@@ -17,6 +17,7 @@ export interface ClientInfoServiceInterface {
   clientInfo: (
     sessionId: string,
     clientSessionId: string,
-    sourceIp: string
+    sourceIp: string,
+    persistentSessionId: string
```
JHjava, line 21: Does this need to be marked as optional for the types as it might not always exist? `persistentSessionId?: string`

JHjava (approved):

PR #471, opened 2022-01-19
AUTH-1277: add healthcheck endpoint to frontend application

dbes-gds:
## What?

Add healthcheck endpoint to frontend management application.
The endpoint returns a simple '200 OK'

## Why?

ALB and Fargate requires a healthcheck endpoint to know whether the app has started.

## Related PRs

https://github.com/alphagov/di-authentication-account-management/pull/304

### DIFF
--- a/src/app.constants.ts
+++ b/src/app.constants.ts
@@ -41,6 +41,7 @@ export const PATH_NAMES = {
   CONTACT_US: "/contact-us",
   CONTACT_US_SUBMIT_SUCCESS: "/contact-us-submit-success",
   PROVE_IDENTITY: "/prove-identity",
+  HEALTHCHECK: "/healthcheck",
 };
 
 export const HTTP_STATUS_CODES = {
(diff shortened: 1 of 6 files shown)


src/app.constants.ts
```diff
@@ -41,6 +41,7 @@ export const PATH_NAMES = {
   CONTACT_US: "/contact-us",
   CONTACT_US_SUBMIT_SUCCESS: "/contact-us-submit-success",
   PROVE_IDENTITY: "/prove-identity",
+  HEALTHCHECK: { url: "/healthcheck" },
```
sanisc, line 44: you dont need this as a object keep it consistent with the others

src/app.ts
```diff
@@ -144,6 +145,7 @@ function createApp(): express.Application {
   app.use(cookieParser());
   app.use(csurf({ cookie: getCSRFCookieOptions(isProduction) }));
 
+  app.use(healthcheckRouter);
```
sanisc, line 148: you should be able to add the route in its normal place as the app structure is different here.

sanisc (changes requested):

sanisc (approved):

PR #524, opened 2022-02-25
AUTH-1385 - Add referer to Zendesk api 

JHjava:
## What?

- Add the referer to the Zendesk body when it is present 
- Make the back button dynamic on the contact-us page so it can return the user to where they previously were 
- When theme isn't present in the requests to contact-us-further-information or contact-us-questions then redirect to contact-us.

## Why?

- So we can have a better idea of where users are facing issues 
- So user can return to the previous page if they were stuck 
- Because if a user goes to them 2 pages without a theme they can't do anything so we should redirect them to the contact-us page

### DIFF
--- a/src/components/contact-us/contact-us-service.ts
+++ b/src/components/contact-us/contact-us-service.ts
@@ -80,10 +80,10 @@ export function contactUsService(
     htmlBody.push(`<p>${optionalData.sessionId}</p>`);
 
     htmlBody.push(`<span>[From page]</span>`);
-    if (!referer) {
-      htmlBody.push(`<p>undefined</p>`);
-    } else {
+    if (referer) {
       htmlBody.push(`<p>${referer}</p>`);
+    } else {
+      htmlBody.push(`<p>Unable to capture referer</p>`);
     }
 
     htmlBody.push(`<span>[User Agent]</span>`);
--- a/src/components/contact-us/contact-us-service.ts
+++ b/src/components/contact-us/contact-us-service.ts
@@ -25,7 +25,8 @@ export function contactUsService(
             contactForm.descriptions,
             contactForm.optionalData,
             contactForm.questions,
-            contactForm.themeQuestions
+            contactForm.themeQuestions,
+            contactForm.referer
           ),
         },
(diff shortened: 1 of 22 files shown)


src/components/contact-us/contact-us-service.ts
```diff
@@ -77,6 +79,13 @@ export function contactUsService(
     htmlBody.push(`<span>[Session ID]</span>`);
     htmlBody.push(`<p>${optionalData.sessionId}</p>`);
 
+    htmlBody.push(`<span>[From page]</span>`);
+    if (!referer) {
```
sanisc, line 83: i would flip this round to be 

```
if (referer) {

htmlBody.push(`<p>${referer}</p>`);
    } else {
      htmlBody.push(`<p>undefined</p>`);
    }
```

As the message is for zendesk users i would change undefined to something a bit easier to understand like unable to capture referrer

src/components/contact-us/tests/contact-us-questions-controller.test.ts
```diff
@@ -170,65 +195,75 @@ describe("contact us questions controller", () => {
...
       req.headers.referer = "/contact-us-further-information";
+      req.query.fromPage = "https://gov.uk/sign-in";
       contactUsQuestionsGet(req as Request, res as Response);
 
       expect(res.render).to.have.calledWith("contact-us/questions/index.njk", {
         theme: "account_creation",
         subtheme: "technical_error",
         backurl: "/contact-us-further-information",
         pageTitleHeading: "pages.contactUsQuestions.technicalError.title",
+        fromPage: "https://gov.uk/sign-in",
```
sanisc, line 251: i would declare this as a variable at the top as its a constant.

src/components/contact-us/contact-us-service.ts
JHjava: Updated

src/components/contact-us/tests/contact-us-questions-controller.test.ts
JHjava: Good shout. Done

sanisc (approved):

PR #526, opened 2022-02-28
AUTH-1389: Remove PaaS Redis resource

chrisclayson:
## What?

- Remove the Redis instance from PaaS

## Why?

We are no longer running Frontend in PaaS so we do not need the Redis instance

*N.B. the CloudFoundry Terraform provider will be removed in a follow-up PR, once this PR has been merged and Terraform has destroyed the Redis resource. The follow up PR will also remove any, now, defunct variables and update the deployment task appropriately*

### DIFF
--- a/ci/terraform/redis.tf
+++ b/ci/terraform/redis.tf
@@ -2,16 +2,6 @@ locals {
   redis_port_number = 6379
 }
 
-data "cloudfoundry_service" "redis" {
-  name = "redis"
-}
-
-resource "cloudfoundry_service_instance" "redis" {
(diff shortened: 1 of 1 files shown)


alan-gds:
Should the `redis_service_plan` be removed now too, since it's not used anywhere else?

chrisclayson:
> Should the `redis_service_plan` be removed now too, since it's not used anywhere else?

Good spot ... I'll do that in the follow up PR that removes the CloudFoundry provider and updates the deployment task.

alan-gds (approved):

PR #543, opened 2022-03-11
Change to post form on account created page

sanisc:
## What?

Change account created page to be a form post.

## Why?

This is to stop issues with users clicking the link more than once which causes error's downstream.


JHjava (approved):
👍

JHjava (approved):

PR #615, opened 2022-06-14
AUT-416: Add 'Choose how to get security codes' screen.

dbes-gds:
## What?

Add 'Choose how to get security codes' screen.

- Add the new screen.
- Add new environment config flag to switch mfa options on and off.
- Screen is currently switched-off in all environments, switch it on locally to test.
- Update state machine to navigate to the new page if the option is on.
- Currently goes to the 'Enter Phone Number' screen regardless of the option chosen as the 'Set up an authenticator app' screen is not yet available.
- There are separate tickets to update the create account drop-out flows, so the state machine has not been changed for this scenario.

## Why?

Users will be given a choice between using SMS and an authenticator app as a second factor.

### DIFF
--- a/src/app.ts
+++ b/src/app.ts
@@ -23,6 +23,7 @@ import {
   getRedisPort,
   getSessionExpiry,
   getSessionSecret,
+  supportMFAOptions,
 } from "./config";
 import { logErrorMiddleware } from "./middleware/log-error-middleware";
 import { enterEmailRouter } from "./components/enter-email/enter-email-routes";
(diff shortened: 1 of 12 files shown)


src/components/multi-factor-authentication/get-security-codes.njk
```diff
@@ -0,0 +1,62 @@
+{% extends "common/layout/base.njk" %}
+{% from "govuk/components/button/macro.njk" import govukButton %}
+{% from "govuk/components/inset-text/macro.njk" import govukInsetText %}
+{% from "govuk/components/radios/macro.njk" import govukRadios %}
+{% from "govuk/components/details/macro.njk" import govukDetails %}
```
sanisc, line 5: rename this file to index.njk its how all the other components are.

src/app.ts
```diff
@@ -83,6 +85,9 @@ function registerRoutes(app: express.Application, appEnvIsProduction: boolean) {
   app.use(resetPasswordCheckEmailRouter);
   app.use(checkYourEmailRouter);
   app.use(createPasswordRouter);
+  if (supportMFAOptions()) {
+    app.use(multiFactorAuthenticationRouter);
```
sanisc, line 89: i would think about the name of this component as mfa doesn't really mean much. It is more about selecting your mfa option. Some ideas:
select-mfa-options
choose-mfa-option

src/components/multi-factor-authentication/tests/multi-factor-authentication-controller.test.ts
```diff
@@ -0,0 +1,67 @@
+import { expect } from "chai";
```
sanisc, line 1: probably worth adding a integration test to verify the validation works correctly and state machine.

sanisc (approved):

PR #620, opened 2022-06-17
BAU: Temporarily reduce the amount of logs we get from AWS

mrwilson:
Does exactly what it says on the tin


chrisclayson (approved):

chrisclayson (changes requested):
Actually, I think this is the wrong thing.

chrisclayson (approved):
Yup, this is neater 👍🏻

PR #640, opened 2022-06-24
AUT-460 - Update arial label to be consistent with screen

JHjava:
## What?

- Update arial label to be consistent with screen

## Why?

- The visual label is not the same as the source code which can cause issues with a screen reader. Update the source code so it is consistent and will work with screen readers

### DIFF
--- a/src/components/common/layout/banner.njk
+++ b/src/components/common/layout/banner.njk
@@ -3,11 +3,6 @@
 <p class="govuk-body">{{ 'general.cookie.cookieBanner.paragraph2' | translate }}</p>
 {% endset %}
 
-{% set html %}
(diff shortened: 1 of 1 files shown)


src/components/common/layout/banner.njk
```diff
@@ -57,7 +57,7 @@
           type: "button",
           classes:"cookie-hide-button",
           attributes: {
-            "aria-label": "Hide cookie banner button"
+            "aria-label": "Hide this message"
```
sanisc, line 60: this page has some issues that the text is not using translations like it should

sanisc (changes requested):

PR #641, opened 2022-06-24
AUT 440: Add 'Enter the security code shown in your authenticator app' screen for sign in journey

dbes-gds:
## What?

Add 'Enter the security code shown in your authenticator app' screen for sign in journey.

The option will only be displayed if SUPPORT_MFA_OPTIONS is turned on, and if the user has auth app credentials in the database, otherwise it will default to SMS.

## Why?

Allows the user to sign in with an authenticator app if they have set one up, rather than using SMS as a second factor.

### DIFF
--- a/src/components/enter-password/enter-password-controller.ts
+++ b/src/components/enter-password/enter-password-controller.ts
@@ -15,6 +15,7 @@ import {
 } from "../common/constants";
 import { BadRequestError } from "../../utils/error";
 import { USER_JOURNEY_EVENTS } from "../common/state-machine/state-machine";
+import { supportMFAOptions } from "../../config";
 
 const ENTER_PASSWORD_TEMPLATE = "enter-password/index.njk";
 const ENTER_PASSWORD_VALIDATION_KEY =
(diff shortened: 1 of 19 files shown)


src/locales/cy/translation.json
```diff
@@ -1420,6 +1425,20 @@
...
+    "enterAuthenticatorAppCode": {
+      "title": "Enter the security code shown in your authenticator app",
+      "header": "Enter the security code shown in your authenticator app",
+      "code": {
+        "label": "Security code",
+        "validationError": {
+          "required": "Enter the security code",
+          "maxLength": "Enter the security code using only 6 digits",
+          "minLength": "Enter the security code using only 6 digits",
+          "invalidFormat": "Enter the security code using only 6 digits",
```
JHjava, line 1438: Do we need all 3 of these?

src/components/enter-password/enter-password-controller.ts
```diff
@@ -128,6 +128,9 @@ export function enterPasswordPost(
           isPhoneNumberVerified: userLogin.data.phoneNumberVerified,
```
JHjava, line 128: Doesn't have to be done now but after we should tidy up and remove the `phoneNumberVerified` from the login response and just use `mfaMethodVerified`

JHjava (approved):

PR #673, opened 2022-07-12
AUT-497: Add unit tests for di cookies policy page

dbes-gds:
## What?

Add unit tests for di cookies policy page.

## Why?

Catching up with the unit tests as the page was needed asap.

## Related PRs

#671 
#670

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,6 +25,7 @@
     "test:audit": "yarn audit",
     "test:integration": "rm -rf test/coverage && NODE_ENV=development nyc mocha -r dotenv/config  \"src/**/*-integration.test.ts\"",
     "test:unit": "rm -rf test/coverage && NODE_ENV=development nyc mocha --exclude \"src/**/*-integration.test.ts\"  \"test/unit/**/*.test.ts\" --recursive \"src/**/*.test.ts\"",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -25,6 +25,7 @@
     "test:audit": "yarn audit",
     "test:integration": "rm -rf test/coverage && NODE_ENV=development nyc mocha -r dotenv/config  \"src/**/*-integration.test.ts\"",
     "test:unit": "rm -rf test/coverage && NODE_ENV=development nyc mocha --exclude \"src/**/*-integration.test.ts\"  \"test/unit/**/*.test.ts\" --recursive \"src/**/*.test.ts\"",
+    "test:u1": "rm -rf test/coverage && NODE_ENV=development nyc mocha --exclude \"src/**/*-integration.test.ts\" --recursive \"src/**/your-test*.test.ts\"",
```
JHjava, line 28: what's this for?

dbes-gds: That's an additional task for when you only want to run one unit test rather than the whole lot, thought it might be useful rather than changing the existing ones to do the same thing.

JY0280: Thanks for asking James and for the response Daniel, I didn't know what this was for either.

JY0280 (approved):

PR #675, opened 2022-07-12
AUT-537: Add length check in auth app code validation

DTaylorGDS:
## What?

- Add length check in auth app code validation
- Add new validation message which displays per below:

image

## Why?

To avoid hitting back end with codes we know are invalid

## Related PRs

[Please include links to PRs in other repositories relevant to this PR.
Delete this section if not needed.](https://github.com/alphagov/di-authentication-api/pull/2131) - back end PR which restricts code length from that end


JHjava:
Looks good @DTaylorGDS. Did Kim provide the content for the error message?

PR #680, opened 2022-07-14
AUT-534: Redisplay auth app qr code and secret after validation error

dbes-gds:
## What?

Redisplay auth app QR code and secret after validation error.

Add an additional optional parameter to 'validateBodyMiddleware' and 'renderBadRequest' to allow for locals to be passed back to the page after a validation error.

## Why?

When the user entered an invalid or incorrect code the page was being redisplayed without the QR code and secret.

### DIFF
--- a/src/components/setup-authenticator-app/index.njk
+++ b/src/components/setup-authenticator-app/index.njk
@@ -26,12 +26,12 @@
 <p class="govuk-body">{{'pages.setupAuthenticatorApp.step2' | translate }}</p>
 
 <p class="govuk-body">
-  <img src="{{qrCode}}" alt="QR Code Image">
+  <img id="qr-code" src="{{qrCode}}" alt="QR Code Image">
 </p>
 
 <p class="govuk-body">
     {{ 'pages.setupAuthenticatorApp.secretKeyLabelText' | translate }}
-    <span class="govuk-body govuk-!-font-weight-bold">{{secretKey}}</span>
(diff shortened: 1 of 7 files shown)


src/components/setup-authenticator-app/index.njk
```diff
@@ -47,7 +48,7 @@
   hint: {
       text: 'pages.setupAuthenticatorApp.code.hintText' | translate
   },
-  classes: "govuk-!-width-two-thirds govuk-!-font-weight-bold",
+  classes: "govuk-input--width-10 govuk-!-font-weight-bold",
```
DTaylorGDS, line 51: Assume this is harmonised with other places now (I quickly checked `enter-authenticator-app-code` view, which is the only other place I know this pops up)

src/types.ts
```diff
@@ -45,7 +45,7 @@ export interface UserSession {
   docCheckingAppUser?: boolean;
   identityProcessCheckStart?: number;
   authAppSecret?: string;
-  cookies_referer?: string;
```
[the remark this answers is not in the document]
dbes-gds: yes that's actually tidying up a thing from yesterday to do with cookies, needed to remove this from the session separately for ZDD.

src/components/setup-authenticator-app/index.njk
dbes-gds: We need to discuss this with Content.

PR #683, opened 2022-07-18
[DO NOT MERGE] AUT-535: Add option to go back from SMS MFA code entry screen

DTaylorGDS:
One of the links currently in this PR will need to change - the 'resend code' link on the SMS MFA journey. This has not yet been implemented, but is expected to go live before the auth app journey.

## What?

- Amend state machine to make it possible to return to MFA type selection screen from phone number set-up screen i.e. allowing auth app selection instead
- Add links in details element to allow change of phone number or change of MFA type selection
- Amend content to match Figma
-- Includes changing enter your mobile phone number page not to differentiate between returning and non-returning user
-- Includes changing enter your mobile phone number page to include paragraph on You need to add a UK mobile phone number to your GOV.UK account to not display when international phone numbers are enabled

## Why?

So that a user is able to reverse MFA method selection up to the point that it is set up e.g. because they have a problem receiving security code via SMS or because their auth app cannot successfully scan the QR code.

### DIFF
--- a/src/components/check-your-phone/index.njk
+++ b/src/components/check-your-phone/index.njk
@@ -3,27 +3,32 @@
 {% from "govuk/components/button/macro.njk" import govukButton %}
 {% from "govuk/components/details/macro.njk" import govukDetails %}
 {% from "govuk/components/error-summary/macro.njk" import govukErrorSummary %}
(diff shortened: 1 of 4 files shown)


DTaylorGDS:
Picked up in other ticket work

PR #718, opened 2022-08-15
AUT-627 - Remove supportMFAOptions flag from frontend code base

JHjava:
## What?

- This required a few minor tweaks to the state-machine.ts to ensure that the users who dropped off would go to the GET_SECURITY_CODES screen instead of the CREATE_ACCOUNT_ENTER_PHONE_NUMBER screen.
- Remove content that is no longer required
- There will be further changes to remove the SUPPORT_MFA_OPTIONS from terraform and other files.

## Why?

- Auth Apps has now been deployed in every environment so now the feature flag can be removed.

### DIFF
--- a/src/locales/en/translation.json
+++ b/src/locales/en/translation.json
@@ -123,8 +123,7 @@
       },
       "paragraph": "You’ll need:",
       "bullet1": "an email address",
-      "bullet2": "a UK mobile phone number",
-      "bullet2AuthApps": "a way to get security codes - this can be a UK mobile phone number or an authenticator app",
+      "bullet2": "a way to get security codes - this can be a UK mobile phone number or an authenticator app",
       "bullet2IntNumbers": "a way to get security codes - this can be a mobile phone number or an authenticator app",
       "insetAlternativeLanguage": {
         "paragraph1": "The GOV.UK account is also available ",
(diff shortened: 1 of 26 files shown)


src/locales/en/translation.json
```diff
@@ -113,8 +113,7 @@
       },
       "paragraph": "You’ll need:",
       "bullet1": "an email address",
-      "bullet2": "a UK mobile phone number",
-      "bullet2AuthApps": "a way to get security codes - this can be a UK mobile phone number or an authenticator app",
```
gtvj, line 117: Do these two values also need to be removed from `/cy/translation.json`?

gtvj (commented):
Hi @DTaylorGDS - I've reviewed what I can for this PR (the backend stuff is beyond my circle of competence, I'm afraid). The frontend looks good but I've just added a small comment about some values in the welsh translation file.

dev-app.js
```diff
@@ -10,6 +10,8 @@ const port = 2000;
 const AUTHORIZE_REQUEST =
   process.env.API_BASE_URL +
   "/authorize?" +
+  // Uncomment the following line to enable P2 Level of Confidence (where back-end environment in .env supports this)
```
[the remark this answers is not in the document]
DTaylorGDS: Thanks, Chris - fair point; removed and will PR separately

src/locales/en/translation.json
DTaylorGDS: Thanks Gwyn - you're right

gtvj (approved):
Hi @DTaylorGDS - from a frontend perspective the PR looks good to me.

PR #727, opened 2022-08-18
AUT-599: Update content for get security code resend code

gtvj:
## What?

- [ ] Updates content for the forgotten password resend code page (See "Accessing the page" below)
- [ ] Creates a new Jira ticket to capture the need for the page to have its own entries within `translation.json` (See "Borrowing translations" below)

### Accessing the page
This page is accessed by: visiting `/sign-in-or-create` and clicking "Sign in"; then entering an existing email address and clicking "Continue"; then clicking "I've forgotten my password"; then expanding "Problems with the code" and clicking "Send the code again"

### Borrowing translations
This PR is being raised during a freeze to the content of `translation.json` (while the team awaits the translators to provide a Welsh version). For this reason the template has borrowed existing values from other components. Once the freeze has lifted we'll need to go back and untangle things.  

## Why?

Part of AUT-599

## Related PRs

#722 #717 #715


gtvj:
I've updated the code to not amend the `translation.json` because we are awaiting the translated text from the translators. Rather than overwriting the previous commit I've reverted so that the initial commit can act as a reference for where the translations might be put. This might not be necessary but it seemed a reasonable thing to do. Happy to discuss though.

gtvj:
This is very odd. While GitHub says I merged the PR into `main` I haven't (there are no commit to merge on this branch). What I did was, following the discussion this morning which revealed the changes I made yesterday were not necessary I: reset all commits that had been made on this branch; pushed the branch with no commits. This seems to have caused GitHub to think I merged (presumably because the branch and `main` have the same tip). 

To avoid any further confusion I'll move the work to another branch and raise a separate PR.

PR #746, opened 2022-09-02
Use smart quotes in account confirmation copy

benilovj:
## What?

Smart quotes in confirmation message, for typographical correctness:

image

## Why?

Please include reason for the change and any other relevant context.

### DIFF
--- a/src/locales/en/translation.json
+++ b/src/locales/en/translation.json
@@ -345,8 +345,8 @@
       }
     },
     "accountCreated": {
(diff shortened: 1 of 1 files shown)


gtvj:
Hi @benilovj, thank you for bringing this to our attention. The team discussed the smart quotes issue and agreed that we should use them. Having reviewed the code we found several instances, all of which have now been replaced. We’ve also introduced a testing mechanism to prevent non-smart quotes from returning which we’ll be sharing with other teams.

PR #756, opened 2022-09-08
AUT-712: Fix radio focus from error summary href (contact-us)

DTaylorGDS:
## What?

- Due to clash between IDs (indirect result of `idPrefix` macro config) and `name`, the `<a>` in the error summary box when clicked does not focus on first radio button.
- Removing `idPrefix` removes the clash, and allows the focus to work as intended.
- Note: this bug still applies on other pages, but ticket relates only to this one page

image

## Why?

- To improve accessibility

## Related PRs

- I started a draft PR, since closed, here: https://github.com/alphagov/di-authentication-frontend/pull/755
- This was on the erroneous understanding that removing `idPrefix` would mean radio buttons would not have unique IDs. Whilst this turned out to be wrong, since `id` is still uniquely populated using the `name` value as the base `id` - hence closing the PR - that approach is still possible if there is a good reason to keep `id` names the same as they previously were.

### DIFF
--- a/src/components/contact-us/index-public-contact-us.njk
+++ b/src/components/contact-us/index-public-contact-us.njk
@@ -37,7 +37,6 @@
 <input type="hidden" name="referer" value="{{referer}}"/>
 
     {{ govukRadios({
-        idPrefix: "contact-us",
         name: "theme",
         fieldset: {
             legend: {


gtvj:
Thanks Dom. I've checked this locally and can see the change has addressed the issue.

gtvj (approved):

PR #763, opened 2022-09-09
AUT-691: Configure cookie domain for lng cookie

dbes-gds:
## What?

Configure cookie domain for lng cookie.

## Why?

As the lng cookie is currently set to the default domain for each app it is not visible across different apps, such as Frontend and Account Management.  This means that when the user chooses a language in Frontend it is not reflected in Account Management as AM sets its own cookie.

Instead, set it to the service domain so that the cookie can be seen across the service.

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -108,3 +108,7 @@ export function getZendeskGroupIdPublic(): number {
 export function getAnalyticsCookieDomain(): string {
   return process.env.ANALYTICS_COOKIE_DOMAIN;
 }
+
+export function getServiceDomain(): string {
+  return process.env.SERVICE_DOMAIN;
+}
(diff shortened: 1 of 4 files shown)


src/config.ts
```diff
@@ -108,3 +108,7 @@ export function getZendeskGroupIdPublic(): number {
 export function getAnalyticsCookieDomain(): string {
   return process.env.ANALYTICS_COOKIE_DOMAIN;
 }
+
+export function getServiceDomain(): string {
+  return process.env.SERVICE_DOMAIN;
```
DTaylorGDS, line 113: Is there any sensible default value here? Or does `undefined` work okay?

DTaylorGDS (approved):
One non-blocking question

src/config.ts
dbes-gds: good question, let me add an additional check in the next PR

PR #765, opened 2022-09-14
AUT-691: i18next cookie tweaks

dbes-gds:
## What?

i18next cookie tweaks.

- Remove use of SameSite for the lng cookie.
- Remove lng parameter from configuration.
- Update the local stub to enable testing of the ui_locales parameter.

## Why?

Fixes an issue where the lng cookie set by the backend in /authorize was not being respected by the frontend, so service language preferences were not being respected.

- lng parameter overrides language detection and is not required when detection is in use.  The fallback language (en) still applies if no language is set.

## Related PRs

#763 
https://github.com/alphagov/di-authentication-account-management/pull/602


README.md
```diff
@@ -26,10 +26,13 @@ SESSION_EXPIRY=30000
 SESSION_SECRET=123456
 API_KEY=
 ANALYTICS_COOKIE_DOMAIN=localhost
+SERVICE_DOMAIN=localhost
 ```
 
 You can find the `API_BASE_URL` in [Concourse](https://cd.gds-reliability.engineering/teams/verify/pipelines/di-authentication-deployment) under the outputs within the deploy-oidc-api-{environment} job, where {environment} is the name of the environment you want to connect to.
 
+`UI_LOCALES` can be used be the stub to request specific locales when authorising.  Only 'en' and 'cy' are supported.
```
gtvj, line 34: Just wondering if there's a typo here with 'be' repeated.

dev-app.js
```diff
@@ -55,3 +67,23 @@ app.listen(port, () => {
   console.log("TEST APP TO REDIRECT FOR NEW SESSION : DEV ONLY");
   console.log(`RUNNING ON http://localhost:${port}`);
 });
+
+function getCookieValue(cookie, cookieName) {
+  var value;
```
gtvj, line 72: Just a minor observation here but wondering if - since `let` is used elsewhere in similar contexts within the file - it might be preferable to use it here too.

gtvj (approved):
Hi @dbes-gds - looks good. Just a couple of minor observations that you can make a judgment on.

PR #787, opened 2022-10-07
ACI-7: Don't ask for email again after max verify code retries

dbes-gds:
## What?

Do not ask a new user to re-enter email when they request another code having exceeded the maximum number of retries when entering an email validation code.

To be deployed in conjunction with a server side change that does not block the email when the maximum number of retries has been reached, and which allows a brand new code to be sent to the email address.

Before:

1. User enters incorrect six digit email security code 6 times during account creation
2. User sees screen telling them they need to wait 15 mins to get a new code.
3. User clicks 'get a new code'
4. User enters same email address (again)
5. User sees screen telling them they need to wait 15 mins to get a new code.
6. User waits 15 mins (30s in build).
7. User clicks 'get a new code' (or abandons and starts again)
8. User enters same email address (again)
9. User user selects 'continue'.
10. User is sent a security code and is redirected to ‘check your email’ screen
11. User can enter the code and continue their journey

After (both frontend and server side changes):

1. User enters incorrect six digit email security code 6 times during account creation
2. User sees screen telling them they need to get a new code (You entered the wrong security code too many times)
3. User selects ‘get new code’
4. User is taken to the 'Get security code' page where they select 'Get security code'
5. User is sent a new security code and is redirected to ‘check your email’ screen, without having to enter their email address again.
6. User can enter the code and continue their journey

Note that if the user abandons the journey before confirming their email address, they will be asked to enter their email address when they try again.

## Why?

No reason to block the account if the account does not already exist, this just means the user has to wait 15 mins before they can create the account.
Make it easier for the user by not asking them to enter their email again.

## Related PRs

https://github.com/alphagov/di-authentication-api/pull/2412
https://github.com/alphagov/di-authentication-acceptance-tests/pull/149

### DIFF
--- a/src/components/common/send-notification/send-notification-service.ts
+++ b/src/components/common/send-notification/send-notification-service.ts
@@ -19,7 +19,8 @@ export function sendNotificationService(
     sourceIp: string,
     persistentSessionId: string,
     userLanguage: string,
-    phoneNumber?: string
+    phoneNumber?: string,
+    requestNewCode?: boolean
   ): Promise<ApiResponseResult<DefaultApiResponse>> {
     const payload: any = {
       email,
@@ -28,6 +29,10 @@ export function sendNotificationService(
 
     if (phoneNumber) {
       payload.phoneNumber = phoneNumber;
+    }
+
+    if (requestNewCode) {
+      payload.requestNewCode = requestNewCode;
     }
 
     const response = await axios.client.post<DefaultApiResponse>(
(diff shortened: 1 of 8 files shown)


src/components/common/send-notification/send-notification-service.ts
```diff
@@ -30,6 +31,10 @@ export function sendNotificationService(
       payload.phoneNumber = phoneNumber;
     }
 
+    if (requestNewCode) {
```
DTaylorGDS, line 34: Not a huge quibble, but I guess this could equally well be `payload.requestNewCode = requestNewCode || false` and avoid the `if` statement altogether, or `requestNewCode` could have a default value of false. It makes little practical difference, but in some way might be nicer always to send this field to the back-end? Don't feel strongly, and there may be reasons not to do this, but that'd be my line of thinking

DTaylorGDS (approved):

PR #815, opened 2022-10-26
BAU: Add Welsh version of untranslated paragraph

wilsond-gds:
## What?

Update `translation.json` file.

## Why?

Welsh translation file included untranslated English content.


DTaylorGDS (approved):

src/locales/cy/translation.json
```diff
@@ -1551,7 +1551,7 @@
         "paragraph2": "Fe allwch wedyn naill ai:",
         "listItem2_1": "llwytho llun o’ch ID gyda llun gan ddefnyddio’r ap <span lang=\"en\">GOV.UK ID Check</span> (dim ond yn Saesneg y mae’r ap ar gael ar hyn o bryd)",
```
DTaylorGDS, line 1552: What is the thinking here? `en` usually associated with English and `cy` with Welsh (or vice versa if it's about flipping to the opposite). Either way seems a bit strange...?

wilsond-gds: hello Dom - the span is an accessibility feature to programmatically inform browsers that there is a phrase in a foreign language within the text - in this case the English app name in Welsh content - suggested by @gtvj

PR #826, opened 2022-11-04
AUT-389: Create new page to indicate user account is locked

ayoshebby:
## What?

Create a new page with content telling a user they aren't able to log in and should wait 15 mins as they have tried to sign in multiple times and their account is now locked.

## Why?

The error page will help the user understand what has gone wrong and how the user can recover from the error.

## Related PRs

[Frontend changes](https://github.com/alphagov/di-authentication-api/pull/2484)


two-first-names (approved):

JHjava (approved):
Looks good 👍  Might have to pause the pipeline when both PRs are merged to test in build

PR #827, opened 2022-11-04
PYIC-2135 Update content on no-photo-id unhappy path

wilsond-gds:
## What?

One value in `en/translation.json` has been updated.

## Why?

This change removes misleading content for the user.

### DIFF
--- a/src/locales/en/translation.json
+++ b/src/locales/en/translation.json
@@ -1649,7 +1649,7 @@
       },
       "section2": {
         "subHeading" : "What you can do",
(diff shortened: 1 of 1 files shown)


src/locales/en/translation.json
```diff
@@ -1649,7 +1649,7 @@
       },
       "section2": {
         "subHeading" : "What you can do",
-        "paragraph1" : "If your passport has expired, you can still use it to prove your identity up until 18 months after the expiry date."
+        "paragraph1" : "Continue to the service you were trying to use and look for other ways to prove your identity."
```
gtvj, line 1652: Having just searched my development machine I can see the new content is already present in the Welsh translation file

PR #836, opened 2022-11-15
ACI-23: Add translation.json for new error screen

DTaylorGDS:
## What?

- Add translation.json for new error screen
- Screen will display when a user attempts to view a page that requires a session directly via the URL
- Note: CY text is English placeholder for now. Will replace with translation at earliest opportunity.

## Why?

- Needed for a new screen (PR for that to follow later)


DTaylorGDS:
Wrong branch

PR #841, opened 2022-11-18
ACI-23: Add mid-journey session error variant

DTaylorGDS:
## What?

- Add new screen when a user attempts to access a screen requiring an active session directly
- This applies for example when someone bookmarks a page or attempts to follow a link from an email

New screen:
image

Existing screen:
image

## Why?
- Prior to this PR, absence of a session would always result in a screen saying that the page has expired through inactivity
- This would not necessarily be true for all users, and would not be obvious how to recover if the user comes via a bookmark
- In such cases the user is given an opportunity to begin a flow to sign into Account Management
- In all other cases where no session is found, the existing screen about session timeout will still be shown
- There are cases where the Referrer header mechanic will not fully address the problem stated above
- The code is looking for Referrer header to contain gov.uk
- This will not be true on first signin screen if relying party is not using a gov.uk domain
- It will also not be true if the user presses enter from the address bar rather than refreshing
- It will also not be true if the user's browser suppresses Referrer information globally
- This is therefore only intended as an incremental improvement rather than a complete content solution


DTaylorGDS:
Superseded by: https://github.com/alphagov/di-authentication-frontend/pull/957

PR #846, opened 2022-11-22
AUT-12: Add service name to account created page

ayoshebby:
## What?

- Introduce a dynamic element on the ‘you’ve created your account’ screen, to show the user the name of the service the user came from.

## Why?

- To assist the user to understand what has happened, what will happen next and what they need to do.

### DIFF
--- a/src/components/account-created/index.njk
+++ b/src/components/account-created/index.njk
@@ -8,7 +8,7 @@
     
 <h1 class="govuk-heading-l govuk-!-margin-top-0 govuk-!-margin-bottom-3">{{'pages.accountCreated.header' | translate}}</h1>
 
-<p class="govuk-body">{{'pages.accountCreated.text' | translate}}</p>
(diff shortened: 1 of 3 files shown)


src/components/account-created/index.njk
```diff
@@ -8,7 +8,7 @@
     
 <h1 class="govuk-heading-l govuk-!-margin-top-0 govuk-!-margin-bottom-3">{{'pages.accountCreated.header' | translate}}</h1>
 
-<p class="govuk-body">{{'pages.accountCreated.text' | translate}}</p>
+<p class="govuk-body">{{'pages.accountCreated.text' | translate}} {{name}} {{'pages.accountCreated.textEnd' | translate}}</p>
```
dbes-gds, line 11: is 'name' passed in somewhere else?

dbes-gds (approved):

PR #853, opened 2022-11-25
ACI-132: Update Node.js to latest LTS

gtvj:
## What?

Updates Node.js to latest LTS

To achieve this I have:

* I updated all references to Node 16.7.0 to reference 18.12.1
* Deleted containers with `docker compose down` and `docker system prune --all`
* Rebuilt the containers with `docker compose up —build`
* Copied the SHA for 18.12.1 from the build output in Terminal
* Added the SHA to all 18.12.1 references within Dockerfiles
* Deleted and rebuilt the containers against this pin
* Ran a shell from the container with `docker exec -it di-authentication-frontend_di-auth-frontend /bin/sh`
* Ran `yarn upgrade —latest`
* Ran `yarn audit` (which reported 0 vulnerabilities)

## Why?

To benefit from the more recent updates to Node.js


gtvj:
The update from 16.7.0 to 18.12.1 introduced so many dependency issues that a stepped upgrade (from 16 to 17 then 18) might be more manageable. Will look into this in a separate PR.

PR #863, opened 2022-12-05
ACI-69: Remove redundant template

gtvj:
## What?

Removes the `enter-email/index.njk` template.

## Why?

The `enter-email/index.njk` template is redundant with relevant journeys being covered by the `index-create-account.njk` and `index-existing-account.njk` templates.


JHjava:
@gtvj is there any unusued text in this file that now can be removed from the en/cy locales files?

JHjava (approved):

gtvj:
Thanks @JHjava - I've updated the translation files to remove the associated values and updated some references in `enter-email-validation.ts`

JHjava (approved):

PR #872, opened 2022-12-12
BAU: Document TEST_CLIENT_ID in the readme

dbes-gds:
## What?

Document TEST_CLIENT_ID in the readme.

## Why?

This was missing from the sample and the documentation and is required to get a local client working.


README.md
```diff
@@ -31,6 +32,8 @@ SERVICE_DOMAIN=localhost
 
 You can find the `API_BASE_URL` in [Concourse](https://cd.gds-reliability.engineering/teams/verify/pipelines/di-authentication-deployment) under the outputs within the deploy-oidc-api-{environment} job, where {environment} is the name of the environment you want to connect to.
 
+`TEST_CLIENT_ID` must be a client configured in the client registry of the environment you are connecting to.  Look in the `environment-client-registry` DynamoDb table and retrieve the value of the `ClientID` field for the relevant client.
```
philandstuff, line 35: This doesn't help explain which client I should be choosing. I'm guessing I want one for which TestClient = 1?

dbes-gds: Not necessarily, that's a different thing to do with suppressing notifications for synthetic users, best choose the one configured to work with the stub

philandstuff: Also while you're in the client registry you need to ensure your configured redirect_uri matches what's in the client registry

PR #877, opened 2022-12-14
FIREBREAK: Enable frontend of A/B Testing proof of concept

gtvj:
## What?

Enables front-end aspects of an A/B testing proof of concept that works as follows: 

1. Where `landing-controller.ts` encounters `startAuthResponse.data.featureFlags` they are added to the session
2. The presence or absence of these `featureFlags` on the session will determine which template is rendered by `create-password-controller.ts`
3. Each template is rendered with a `data-ab-test` attribute that has a value of either "variant-a" or "variant-b"
4. A new section is added to `translation.json` files for content that is included in A/B tests (the intention here is to allow it to be identified quickly and removed when no longer needed)
5. Where the user has consented to analytics cookies a JavaScript function will run that captures the `path` and variant information and pushes this to `dataLayer`

## Why?

Firebreak proof of concept to explore potential for A/B testing.

## Related PRs

A [corresponding PR](https://github.com/alphagov/di-authentication-api/pull/2610) has been raised for related API changes.

### DIFF
--- a/src/components/create-password/create-password-controller.ts
+++ b/src/components/create-password/create-password-controller.ts
@@ -11,7 +11,10 @@ import {
 } from "../../utils/validation";
 
 export function createPasswordGet(req: Request, res: Response): void {
-  if(req.session.user.featureFlags && req.session.user.featureFlags.updatePasswordHintTextVersion === '2') {
+  if (
+    req.session.user.featureFlags &&
+    req.session.user.featureFlags.updatePasswordHintTextVersion === "2"
+  ) {
     return res.render("create-password/ab-tests/index-variant.njk");
   }
   res.render("create-password/index.njk");
--- a/src/components/create-password/create-password-controller.ts
+++ b/src/components/create-password/create-password-controller.ts
@@ -11,6 +11,9 @@ import {
 } from "../../utils/validation";
 
 export function createPasswordGet(req: Request, res: Response): void {
+  if(req.session.user.abTest) {
+    return res.render("create-password/ab-tests/index-variant.njk");
+  }
   res.render("create-password/index.njk");
 }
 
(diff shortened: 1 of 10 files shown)


src/components/create-password/create-password-controller.ts
```diff
@@ -11,6 +11,12 @@ import {
 } from "../../utils/validation";
 
 export function createPasswordGet(req: Request, res: Response): void {
+  if (
+    req.session.user.featureFlags &&
```
DTaylorGDS, line 15: Is this 'double-boolean' approach preferable to using optional chaining operator (`?.`) for keeping the second check safe against undefined parents?

src/types.ts
```diff
@@ -47,6 +47,7 @@ export interface UserSession {
   authAppSecret?: string;
   authAppQrCodeUrl?: string;
   isPasswordChangeRequired?: boolean;
+  featureFlags?: any;
```
DTaylorGDS, line 50: Will these always be A/B testing feature flags? If yes, wondering if there's any naming confusion between this and the sort of feature flag that is set at app level like 'support international phone numbers'?

src/components/create-password/create-password-controller.ts
```diff
@@ -11,6 +11,12 @@ import {
...
+    req.session.user.featureFlags.updatePasswordHintTextVersion === "2"
+  ) {
+    return res.render("create-password/ab-tests/index-variant.njk");
```
DTaylorGDS, line 18: Definitely a question NOT a suggestion, but do we actually want to display the variant when the user hasn't accepted cookies or does the fact non-cookie-consent users can't be tracked via GA mean that there's no learning benefit for us in showing them the variant, and it's slightly 'safer' for their user experience to show them the 'tried and tested' option.

DTaylorGDS (approved):
Left some comments but nothing that I think absolutely has to change - most questions really.

DTaylorGDS (approved):

src/components/create-password/create-password-controller.ts
dbes-gds: updated

PR #890, opened 2023-01-04
ACI-139: Add character limit to user support forms

gtvj:
## What?

Introduces the `govukCharacterCount` to the following user support forms: 

- [x] (Account creation) You do not have a UK mobile phone number
- [x] (Account creation) You did not get a security code
- [x] (Account creation) The security code did not work
- [x] (Account creation) You had a problem with an authenticator app
- [x] (Account creation) There was a technical problem (for example, the service was unavailable)
- [x] (Account creation) Something else
- [x] (Signing in) You did not get a security code
- [x] (Signing in) The security code did not work
- [x] (Signing in) You’ve forgotten your password
- [x] (Signing in) You’ve changed your phone number or lost your phone
- [x] (Signing in) You’ve been told your account ‘cannot be found’
- [x] (Signing in) There was a technical problem (for example, the service was unavailable)
- [x] (Signing in) Something else
- [x] Another problem using your GOV.UK account
- [x] (ID Check App) You had a problem linking the app to your web browser
- [x] (ID Check App) You had a problem taking a photo of your identity document (for example, passport or driving licence)
- [x] (ID Check App) You had a problem scanning your face
- [x] (ID Check App) There was a technical problem (for example you saw an error message)
- [x] (ID Check App) Something else
- [x] A problem proving your identity
- [x] GOV.UK email subscriptions
- [x] A suggestion or feedback about using your GOV.UK account

## Why?

Analysis has revealed the need to communicate and ensure a limit of 1200 characters on text areas within user support forms. See the Jira ticket for a full description.

## How?

The approach to achieve this has been: 

1. To add a `ZENDESK_FIELD_MAX_LENGTH` constant that represents the maximum length of fields submitted to Zendesk. This introduces more complexity than simply having the value hard-coded in several places but a single source of truth ensures changes to the constant will immediately cascade to the controller, validation logic and messages, and for number localisation.
2. Adds a new validation rules to prevent submissions that exceed this length.
3. Introduces a new object constant `PLACEHOLDER_REPLACEMENTS` to hold patterns that look for placeholders and their replacements
4. Introduces a `replaceErrorMessagePlaceholders()` function that iterates over error messages to match patterns and make replacements found in `PLACEHOLDER_REPLACEMENTS`. These serve to prepare an error list for use within the error summary (because the error summary expects a list it is not possible to use Nunjucks filters here)
5. Implementes the `govukCharacterCount` macro using these mechanisms on pages using it, using Nunjucks filters when instantiating the component (rather than doing the filter upstream)
6. Extracting the existing code that de-duped error messages in `validation.ts` into a function to make the code a bit easier to understand
7. Introduce `validation.test.ts` and new unit tests for these functions

### DIFF
--- a/src/components/contact-us/contact-us-questions-validation.ts
+++ b/src/components/contact-us/contact-us-questions-validation.ts
@@ -221,6 +221,9 @@ export function getErrorMessageForIssueDescriptionWhereLengthExceeded(
   ) {
     return "pages.contactUsQuestions.issueDescriptionErrorMessage.anythingElseTooLongMessage";
   }
+  if (theme === ZENDESK_THEMES.SOMETHING_ELSE) {
+    return "pages.contactUsQuestions.issueDescriptionErrorMessage.entryTooLongMessage";
+  }
 }
 
 export function getErrorMessageForAdditionalDescription(
--- a/src/components/contact-us/contact-us-questions-validation.ts
+++ b/src/components/contact-us/contact-us-questions-validation.ts
@@ -35,7 +35,10 @@ export function validateContactUsQuestionsRequest(): ValidationChainFunc {
       .isLength({ max: ZENDESK_FIELD_MAX_LENGTH })
       .withMessage((value, { req }) => {
         return req.t(
-          "pages.contactUsQuestions.issueDescriptionErrorMessage.entryTooLongMessage",
+          getErrorMessageForIssueDescriptionWhereLengthExceeded(
+            req.body.theme,
+            req.body.subtheme
+          ),
           { value }
         );
       }),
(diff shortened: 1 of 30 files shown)


src/utils/validation.ts
```diff
@@ -18,6 +22,22 @@ export function formatValidationError(
   return error;
 }
 
+export function replaceErrorMessagePlaceholders(errors: {
```
DTaylorGDS, line 25: Again, no opinion, but just something to consider as an alternative. Would it be more straightforward to leverage the replace function as is used here?: https://github.com/alphagov/di-authentication-frontend/blob/ed8c21e5646ace4c45e9d6f78a791e2db715b574/src/components/prove-identity-callback/index.njk#L2

This would work simply by passing the max length variable into the template from the controller and letting the .njk file deal with swapping in the right number...?

DTaylorGDS (approved):

src/components/contact-us/contact-us-questions-validation.ts
```diff
@@ -142,6 +181,90 @@ export function getErrorMessageForIssueDescription(
   }
 }
 
+export function getErrorMessageForIssueDescriptionWhereLengthExceeded(
+  theme: string,
+  subtheme: string
+): string {
+  if (
```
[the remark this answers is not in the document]
gtvj: Hi @DTaylorGDS - I've just pushed a commit to address the Cognitive Complexity issue. Thanks again for making this suggestion. The approach I've opted for is described in the commit message.

src/utils/validation.ts
gtvj: Hi @DTaylorGDS, I got some advice from the Design System Team and have just pushed a commit that addresses this issue. What I've done is

- Use Nunjucks filters when _instantiating_ the `govukCharacterCount` component (rather than doing the filter upstream)
- Creating a new `replaceErrorMessagePlaceholders` function within `validation.ts` that prepares an error list for use within the error summary (because the error summary expects a list it is not possible to use Nunjucks filters here)
- Introducing a `PlaceholderReplacement` interface
- Extracting the existing code that de-duped error messages into a function to make the code a bit easier to understand
- Introduce `validation.test.ts` and new unit tests for these functions

Because these are quite significant changes I've retested every support form (for both missing field and length exceeded errors).

DTaylorGDS (approved):

PR #951, opened 2023-03-08
AUT-1116: Change One Login Sign In link to secondary button

ayoshebby:
## What?

Change current ‘Sign in’ link on the ‘Create a [GOV.UK](http://gov.uk/) One Login or sign in’ page to a secondary button.

## Why?

After seeing an increase in the number of users landing on the `You already have an account` page which was a sign of users mistakenly choosing to create account rather than sign in. To avoid confusion, a secondary button is used as supposed to a text with a link attached.

image

### DIFF
--- a/src/components/sign-in-or-create/index.njk
+++ b/src/components/sign-in-or-create/index.njk
@@ -68,15 +68,19 @@
     text: 'pages.signInOrCreate.createButtonText' | translate,
     value: "create",
     name: "optionSelected",
+    classes: "govuk-!-margin-right-3",
     attributes: {
       "id": "create-account-link"
     }
   }) }}
 
-    <p class="govuk-body">
-      {{ 'pages.signInOrCreate.paragraph2' | translate }}
-      <a href="/sign-in-or-create?redirectPost=true" class="govuk-link" id="sign-in-link">{{ 'pages.signInOrCreate.signInText' | translate }}</a>.
-    </p>
+    {{ govukButton({
+      text: 'pages.signInOrCreate.signInText' | translate,
+      classes: "govuk-button--secondary",
+      attributes: {
+        "id": "sign-in-button"
+      }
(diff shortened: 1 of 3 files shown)


src/locales/en/translation.json
```diff
@@ -156,7 +156,7 @@
         "linkHref": "?lng=cy"
       },
       "paragraph2": "If you already have a GOV.UK One Login you can",
```
DTaylorGDS, line 158: Is this text still being used?

src/locales/en/translation.json
```diff
@@ -156,7 +156,7 @@
...
-      "signInText": "sign in",
+      "signInText": "Sign In",
```
JHjava, line 159: I think the In here should be lower case?

src/components/sign-in-or-create/index.njk
```diff
@@ -68,15 +68,16 @@
...
     }
   }) }}
 
-    <p class="govuk-body">
-      {{ 'pages.signInOrCreate.paragraph2' | translate }}
-      <a href="/sign-in-or-create?redirectPost=true" class="govuk-link" id="sign-in-link">{{ 'pages.signInOrCreate.signInText' | translate }}</a>.
-    </p>
+    {{ govukButton({
+      text: 'pages.signInOrCreate.signInText' | translate,
+      classes: "govuk-button--secondary"
```
CraigRyder, line 79: The previous link for sign-in had an id. If I'm reading this correctly, the new button doesn't appear to have one. Could we have an id for it pls, unless there's a specific reason for not having one? (sign-in-button?)

CraigRyder (commented):
id looks good from my perspective, Ayo. Can't comment on the other bits though i'm afraid

DTaylorGDS (approved):
LGTM

PR #954, opened 2023-03-10
BAU: Additional validation for contact us referer

dbes-gds:
## What?

Additional validation for contact us referer

## Why?

The referer value must either be on an allowlist or a url on our domain.

### DIFF
--- a/src/components/contact-us/contact-us-controller.ts
+++ b/src/components/contact-us/contact-us-controller.ts
@@ -4,12 +4,14 @@ import {
   SUPPORT_TYPE,
   ZENDESK_THEMES,
   ZENDESK_FIELD_MAX_LENGTH,
+  CONTACT_US_REFERER_ALLOWLIST,
 } from "../../app.constants";
 import { contactUsService } from "./contact-us-service";
 import { ContactUsServiceInterface, Questions, ThemeQuestions } from "./types";
 import { ExpressRouteFunc } from "../../types";
 import crypto from "crypto";
 import { logger } from "../../utils/logger";
+import { getServiceDomain } from "../../config";
 
 const themeToPageTitle = {
   [ZENDESK_THEMES.ACCOUNT_NOT_FOUND]:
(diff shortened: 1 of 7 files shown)


src/components/contact-us/contact-us-controller.ts
```diff
@@ -62,27 +65,58 @@ export function contactUsGet(req: Request, res: Response): void {
   if (req.query.supportType === SUPPORT_TYPE.GOV_SERVICE) {
     return res.render("contact-us/index-gov-service-contact-us.njk");
   }
-  let referer = req.headers.referer;
+
+  let referer = validateReferer(req.headers.referer);
```
DTaylorGDS, line 69: I know this doesn't make a practical difference here, but should we prefer `req.get('Referrer')` as a slight guard against the referer/referrer spelling issue?: https://www.geeksforgeeks.org/express-js-req-get-function/

src/components/contact-us/contact-us-controller.ts
```diff
@@ -62,27 +65,58 @@ export function contactUsGet(req: Request, res: Response): void {
...
 
   if (req.query.referer) {
-    referer = req.query.referer as string;
+    referer = validateReferer(req.query.referer as string);
+    logger.info(`referer with referer query param ${referer}`);
   }
 
   if (req.headers.referer && req.headers.referer.includes("referer")) {
```
DTaylorGDS, line 76: Did you try the `req.headers.referer?.includes("referer")` syntax here?

src/components/contact-us/contact-us-controller.ts
```diff
@@ -62,27 +65,58 @@ export function contactUsGet(req: Request, res: Response): void {
...
 }
 
+function validateReferer(referer: string): string {
+  let valid = false;
+  let url;
+  try {
+    if (CONTACT_US_REFERER_ALLOWLIST.includes(referer)) {
+      valid = true;
+    } else {
+      url = new URL(xss(referer));
```
DTaylorGDS, line 101: What is `xss` doing for us here? If it's just stripping HTML tags, does the new URL constructor not also sanitise the input in an equivalent way (genuines question, but I assume HTML tags don't survive URL encoding anyway?)

DTaylorGDS (dismissed):

DTaylorGDS (approved):
LGTM

PR #963, opened 2023-03-23
BAU: startup and pre-commit scripts

dbes-gds:
## What?

Add 'startup' and 'pre-commit' scripts.
Apply formatting changes created by running the 'pre-commit' hooks.

The 'startup' script starts the application locally, either with everything running in docker compose, or in 'local' mode where the frontend application is started on a local machine node instance, with the dependent services still running in docker.

The 'pre-commit' script makes use of the 'pre-commit' hooks created by @chrisclayson.  You may need to run 'brew install pre-commit' for these to work.  The first run might take a while as it has to pull the docker images, so be patient.

Note that there are currently a couple of integration tests that fail on a local machine but run fine in GHA.  These need to be investigated.

## Why?

Provides a quick way to start the app in a local node server.  This is much snappier than running under docker.
Provides a way to run all pre-flight checks without having to rely on github actions.

## Related PRs

https://github.com/alphagov/di-pre-commit-hooks


gtvj:
Hi @dbes-gds - I was just wondering if we might want to add a couple of lines to the `README.md` to explain how to run the app in the different ways - maybe under a new heading alongside the "Running the app in Docker". No worries if not, but I thought I'd mention it as it's the kind of thing I'd find helpful.

dbes-gds:
@gtvj have updated the readme, please take a look

gtvj (approved):

PR #972, opened 2023-03-29
AUT-995: Change content on the /enter-mfa screen accordingly to mfa method type

ayoshebby:
## What?

When a user signs in with 1FA and tries to access another service like account management, they are required to go through 2FA regardless of their authentication level and redirected to `/enter-code` screen.

- Create new field `mfaMethodType` in `UserSession` during sign in journey
- Display new content for users with `MFA_METHOD_TYPE` of `AUTH_APP`

Screenshot 2023-03-30 at 10 21 42

- Display new content for users with `MFA_METHOD_TYPE` of `SMS`
Screenshot 2023-03-30 at 10 22 37

## Why?

So that a user signed in with 1FA understands why they are required to go through 2FA process in order to access the RP's service.


gtvj (commented):
Hi @ayoshebby - just one comment from me about a `<span>` that might not be necessary.

dbes-gds (approved):

PR #973, opened 2023-03-30
AUT-1149: DOM change to satisfy acceptance tests

gtvj:
## What?

Moves `id="secret-key"` to DOM element that contains only the secret key

## Why?

Previous PR had altered the DOM to accommodate the `<span>` elements for formatting and moved the ID. Since merge we've discovered the acceptance tests had relied upon the `id="secret-key"` to verify that secret keys are of the correct length.

## Related PRs
https://github.com/alphagov/di-authentication-frontend/pull/971


gtvj:
On reflection, it seems more sensible to update the acceptance tests so they sum the text to elements that have `secret-key-fragment` (since they directly correspond to the secret key, and only the secret key), rather than introducing additional DOM for this purpose.

Closing this PR.

PR #1004, opened 2023-04-26
AUT-1082: Add variant to MFA selection screen

DTaylorGDS:
## What?
- Add variant to MFA selection screen, to be used in Account Recovery journeys
- Note: this introduces an extra boolean to the User Session, which may cause bugs if we lose track of it. At the moment, I think it is safe, because it has no effect outside of this single page, and once an account recovery journey starts, neither of the other two variants should apply i.e. account creation or account part creation; by definition account (i.e. MFA) recovery would have to happen _after_ either/both of those

The new variant looks like this:
image

Notes:
- Whilst testing, I noticed a bug: all req.body fields were being passed to the render function as a string
- This meant that any hidden inputs (in the end not the approach used here) with value "false" were being interpreted by Nunjucks as truthy so the conditional logic based on them was executing.
- I have changed the validation util to make sure booleans are still booleans when passed back into a template

## Why?
- Part of account recovery journey - as yet under development and not accessible in production

- No changes demonstrated because review will happen at feature level

### DIFF
--- a/src/components/select-mfa-options/index.njk
+++ b/src/components/select-mfa-options/index.njk
@@ -24,7 +24,7 @@
 {% if isAccountPartCreated %}
   {% set radioHeading = 'pages.getSecurityCodes.headerAccountPartCreated' | translate %}
   {% set radioHintText = 'pages.getSecurityCodes.summaryAccountPartCreated' | translate %}
-{% elif isAccountRecovery %}
+{% elif isAccountRecoveryJourney %}
   {% set radioHeading = 'pages.getSecurityCodes.accountRecoveryVariants.header' | translate %}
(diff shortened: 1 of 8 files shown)


src/components/select-mfa-options/index.njk
```diff
@@ -31,6 +36,7 @@
 
 <input type="hidden" name="_csrf" value="{{csrfToken}}"/>
 <input type="hidden" name="isAccountPartCreated" value="{{isAccountPartCreated}}"/>
+<input type="hidden" name="isAccountRecovery" value="{{isAccountRecovery}}"/>
```
dbes-gds, line 39: are these hidden fields used anywhere?

dbes-gds (approved):

PR #1008, opened 2023-05-04
AUT-1242: Refactor VerifyMfaCode interface to use Journey Type parameter

ayoshebby:
## What?

Introduce `JOURNEY_TYPE` parameter for the `VerifyMfaCodeInterface`
- Registration
- Account Recovery
- Sign In

## Why?

In order to redirect to the corresponding feature in the `VerifyMfaCode` lambda.

## Related PRs

[Changes in the Auth API](https://github.com/alphagov/di-authentication-api/pull/2962)

### DIFF
--- a/src/components/common/verify-mfa-code/verify-mfa-code-service.ts
+++ b/src/components/common/verify-mfa-code/verify-mfa-code-service.ts
@@ -1,6 +1,7 @@
 import {
   API_ENDPOINTS,
   HTTP_STATUS_CODES,
+  JOURNEY_TYPE,
   MFA_METHOD_TYPE,
 } from "../../../app.constants";
 import {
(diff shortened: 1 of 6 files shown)


src/components/common/verify-mfa-code/verify-mfa-code-service.ts
```diff
@@ -18,7 +19,7 @@ export function verifyMfaCodeService(
   const verifyMfaCode = async function (
     methodType: MFA_METHOD_TYPE,
     code: string,
-    isRegistration: boolean,
```
JHjava, line 21: Do ensure we have ZDD with this change, we'll have to deploy this first we'll need to keep the `isRegistration` in the request for now. We should be able to remove it after the API is deployed

JHjava (approved):
👍

PR #1019, opened 2023-05-09
AUT-1275: Loosen international phone number validation to include 00 format

DTaylorGDS:
## What?
- Loosen international phone number validation to include 00 prefix format
- For example 0034608453322 would be rejected prior to the change
- After the change it would be transformed into a '+' format number, but accepted in validation

## Implied acceptable numbers for being treated as 'valid international number'
The field in question is the bottom one below:
image

I considered also validating that this field was a non-UK number by changing the following function to:

```
export function containsInternationalMobileNumber(value: string): boolean {
  const formattedNumber = convertInternationalPhoneNumberToE164Format(value);
  return (
      isValidPhoneNumber(formattedNumber) &&
      parsePhoneNumberWithError(formattedNumber).countryCallingCode !== "44"
    );
}
```

This would have eliminated UK-format values like `0044 7777 777 777`. The reason that I did not add this extra check is that I could see unit tests were already accepting E164 format UK numbers:

```
    it("should return true with valid uk mobile in E164 without lib country code", () => {
      expect(containsInternationalMobileNumber("+447911123456")).to.equal(true);
    });
```

It therefore felt most in the spirit of the existing code _not_ to implement an extra check of this kind

## Why?
- User research suggested that supporting the `00` format would be useful - and indeed that users expected to be able to

No change to UI - only validation behaviour

No change to UI/HTML - only validation behaviour

### DIFF
--- a/src/components/enter-phone-number/enter-phone-number-controller.ts
+++ b/src/components/enter-phone-number/enter-phone-number-controller.ts
@@ -10,7 +10,7 @@ import { BadRequestError } from "../../utils/error";
 import { SendNotificationServiceInterface } from "../common/send-notification/types";
 import { sendNotificationService } from "../common/send-notification/send-notification-service";
 import { USER_JOURNEY_EVENTS } from "../common/state-machine/state-machine";
-import { prependInternationalPrefix } from "../../utils/phone-number";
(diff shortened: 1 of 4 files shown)


JHjava (approved):
Is it worth increasing the length we accept from 25 to 26?

JHjava (approved):

PR #1044, opened 2023-05-23
Stop dropping some ALB logs

DTaylorGDS:
## What?
- Stop dropping some ALB logs
- Do not drop when action is BLOCK

## Why?
- Will help diagnostics when ALB is rejecting requests


BeckaL:
Closing as we've now implemented this anyway :-)

PR #1047, opened 2023-05-30
AUT:1281: OTP Block for Account Recovery journey

ayoshebby:
## What?

Ensure user cannot request new code when the user has entered invalid email verification code too many times before the 15 minute blocked period.

- A new lock `wrongCodeEnteredAccountRecoveryLock` is stored in the user session when user is blocked from entering invalid verification code too many times. The API returns error code `1046`

## Why?

To block users from requesting a new code multiple times.

## Related PRs

[Backend change](https://github.com/alphagov/di-authentication-api/pull/3017)

### DIFF
--- a/src/components/resend-email-code/resend-email-code-controller.ts
+++ b/src/components/resend-email-code/resend-email-code-controller.ts
@@ -12,6 +12,19 @@ import { sendNotificationService } from "../common/send-notification/send-notifi
 import xss from "xss";
 
 export function resendEmailCodeGet(req: Request, res: Response): void {
+  if (
+    req.session.user.wrongCodeEnteredAccountRecoveryLock &&
+    new Date().getTime() <
+      new Date(req.session.user.wrongCodeEnteredAccountRecoveryLock).getTime()
+  ) {
+    const newCodeLink = req.query?.isResendCodeRequest
+      ? "/security-code-check-time-limit?isResendCodeRequest=true"
+      : "/security-code-check-time-limit";
+    return res.render("security-code-error/index-wait.njk", {
+      newCodeLink,
+    });
+  }
+
   res.render("resend-email-code/index.njk", {
     emailAddress: req.session.user.email,
     requestNewCode:
--- a/src/components/security-code-error/security-code-error-controller.ts
+++ b/src/components/security-code-error/security-code-error-controller.ts
@@ -6,18 +6,33 @@ import {
 } from "../common/constants";
 import { PATH_NAMES } from "../../app.constants";
 import {
+  getAccountRecoveryCodeEnteredWrongBlockDurationInMinutes,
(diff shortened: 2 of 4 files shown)


src/components/security-code-error/security-code-error-controller.ts
```diff
@@ -18,6 +18,13 @@ export function securityCodeInvalidGet(req: Request, res: Response): void {
       Date.now() + getCodeEnteredWrongBlockDurationInMinutes() * 60000
     ).toUTCString();
   }
+
+  if (req.query.actionType === SecurityCodeErrorType.EmailMaxRetries) {
```
JHjava, line 22: This is going to put on a lock when a user enters more than 5 incorrect codes during registration. Should this SecurityCodeErrorType be `ChangeSecurityCodesEmailMaxRetries` instead?

src/components/security-code-error/security-code-error-controller.ts
```diff
@@ -18,6 +18,13 @@ export function securityCodeInvalidGet(req: Request, res: Response): void {
...
+    req.session.user.codeRequestLock = new Date(
```
JHjava, line 23: How comes you use the codeRequestLock here? I think it should be the`wrongCodeEnteredLock` but you will need to check that using this doesn't impact other parts of the journey such as when a user logs in and wants to enter their MFA code

src/components/resend-email-code/resend-email-code-controller.ts
```diff
@@ -12,6 +12,18 @@ import { sendNotificationService } from "../common/send-notification/send-notifi
 import xss from "xss";
 
 export function resendEmailCodeGet(req: Request, res: Response): void {
+  if (
+    req.session.user.codeRequestLock &&
```
JHjava, line 16: This looks to be the wrong type of lock

JHjava (changes requested):
I think a few changes are needed here. Let me know if you want to discuss

src/components/security-code-error/security-code-error-controller.ts
```diff
@@ -11,13 +11,25 @@ import {
 } from "../../config";
 
 export function securityCodeInvalidGet(req: Request, res: Response): void {
-  const isNotEmailCode =
-    req.query.actionType !== SecurityCodeErrorType.EmailMaxRetries;
+  let isNotEmailCode = false;
```
JHjava, line 14: It's a bit more readable how it was before but perhaps could just change the variable name to `isNotEmailRegsitrationCode`

JHjava (commented):
Small comment but looks good otherwise. Once it is deployed, let's check that the account recovery lock doesn't impact the MFA lock.

JHjava (dismissed):

JHjava (approved):

PR #1057, opened 2023-06-06
Remove Concourse reference from README.md file (AUT-1241)

elenaschwan:
## What and why?

We have migrated off Concourse, which means we need to ensure to not to reference it in our documentation. The README.md file is currently referencing Concourse in the section about setting environment variable, which we can just remove. 


JHjava (approved):

elenaschwan:
Thanks @JHjava! I haven't got write access to the repo, so I think one of you guys will need to merge it. But not sure if we can merge it while we got a failed check...

PR #1062, opened 2023-06-12
AUT-1334: Fix autocomplete for sign in OTP

gtvj:
## What?

A support ticket has submitted by user noting that autocomplete behaviour is not working as expected on this form.

This appears to be due to the presence of `autocomplete="off"` and the absence of the relevant `autocomplete="one-time-code` (which is required under WCAG 2.1 Level AA, 1.3.5 Identify Input Purpose).

I've reviewed the project history to identify why `autocomplete="off"` might have been added to this field and identified a commit and PR but there is no explanation (link to the PR below)

## Why?

Fixes a bug and Level AA WCAG failure. 

## Related PRs
https://github.com/alphagov/di-authentication-frontend/pull/582


JHjava (approved):
Nice. I've had a look and it looks like we've also not got it on in `index-2fa-service-uplift-auth-app.njk` or `index-2fa-service-uplift-mobile-phone.njk` as well as `index.njk` under reset-password-check-email. Do you think we need this changed in these files too?

gtvj:
Thanks for spotting this @JHjava - that's right, we do need it in those too. I also noticed it in another file (`reset-password-check-email/index.njk`) I've just pushed an amended commit to change all four files.

JHjava (approved):

PR #1095, opened 2023-07-11
AUT-1412: Override Password Reset code blocked minutes env in build

ayoshebby:
## What?

- Set `PASSWORD_RESET_CODE_ENTERED_WRONG_BLOCKED_MINUTES` env variable in build to 3000 (30 seconds)
- In order to be consistent with the frontend API

## Why?

Currently in build, when a user is blocked is blocked from entering code, the default elapsed time is set to 15 minutes in build environment. It should rather be set to 30 seconds to be consistent with the Frontend API.

### DIFF
--- a/ci/terraform/build.tfvars
+++ b/ci/terraform/build.tfvars
@@ -5,10 +5,11 @@ frontend_auto_scaling_enabled   = true
 frontend_task_definition_cpu    = 512
 frontend_task_definition_memory = 1024
 
-support_international_numbers        = "1"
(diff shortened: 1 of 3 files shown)


JHjava:
Hi @ayoshebby 

As you've added a value to `build.tfvars`, you will also need to add that terraform variable to `variables.tf` and then make an addition to `config.tf` as well as `ecs.tf`. 

I think it's also worth keeping the value as minutes rather than milliseconds otherwise all the function and environment names will need changing as well.

PR #1112, opened 2023-07-24
BAU - Update account management URL

JHjava:
## What?

- Update account management URL

## Why?

- It should now be home.account.gov.uk instead of account.gov.uk to ensure we take users to account management


BeckaL (approved):
Looks sensible to me!

PR #1114, opened 2023-07-25
AUTH1248: Add claim validation to jwt-service.

gn-GDS:
Add jwt-service tests.
Add AuthorizeRequestPayload interface.

## What?

- Added basic validation of JWE and claims if contained in the query request.

## Why?

- This will remove the dependency of the start API endpoint in the frontend API on client registry (DB) access
- This is a pre-requisite for splitting the orchestration and authentication APIs

## Related PRs
Builds on backend changes - Pass client registry fields to auth front end via JWE: https://github.com/alphagov/di-authentication-api/pull/3159

### DIFF
--- a/src/components/authorize/authorize-controller.ts
+++ b/src/components/authorize/authorize-controller.ts
@@ -59,9 +59,7 @@ export function authorizeGet(
     const claims = await jwtService.getPayloadWithSigCheck(
       authRequestJweDecryptedAsJwt
     );
-    const validateClaims = jwtService.validateClaims(claims);
-
-    validateQueryParamsAgainstClaims(clientId, validateClaims.client_id);
+    jwtService.validateClaims(claims);
 
     const startAuthResponse = await authService.start(
       sessionId,
@@ -169,16 +167,5 @@ function validateQueryParams(clientId: string, responseType: string) {
 
   if (clientId !== expectedClientId) {
     throw new QueryParamsError("Client ID value is incorrect");
-  }
-}
-
-function validateQueryParamsAgainstClaims(
-  requestClientId: string,
-  claimsClientId: string
-) {
-  if (requestClientId !== claimsClientId) {
-    throw new QueryParamsError(
-      "Request Client ID does not equal Claims Client ID"
-    );
   }
 }
--- a/src/components/authorize/authorize-controller.ts
+++ b/src/components/authorize/authorize-controller.ts
@@ -57,10 +57,11 @@ export function authorizeGet(
       encryptedAuthRequestJWE
     );
 
-    const claims = await jwtService.getPayloadWithSigCheck(
+    const claims = await jwtService.getPayloadWithValidation(
       authRequestJweDecryptedAsJwt
     );
-    jwtService.validateClaims(claims);
+
+    jwtService.validateCustomClaims(claims);
 
     const startAuthResponse = await authService.start(
       sessionId,
--- a/src/components/authorize/authorize-controller.ts
+++ b/src/components/authorize/authorize-controller.ts
@@ -24,7 +24,7 @@ import {
 } from "./types";
 import { KmsDecryptionService } from "./kms-decryption-service";
 import { JwtService } from "./jwt-service";
-import { expectedClientId } from "./claims-config";
+import { EXPECTED_CLIENT_ID } from "./claims-config";
 
 function createConsentCookie(
   res: Response,
@@ -167,7 +167,7 @@ function validateQueryParams(clientId: string, responseType: string) {
     throw new QueryParamsError("Response type does not exist");
   }
 
-  if (clientId !== expectedClientId) {
+  if (clientId !== EXPECTED_CLIENT_ID) {
     throw new QueryParamsError("Client ID value is incorrect");
   }
 }
--- a/src/components/authorize/authorize-controller.ts
+++ b/src/components/authorize/authorize-controller.ts
@@ -56,6 +56,7 @@ export function authorizeGet(
     const authRequestJweDecryptedAsJwt = await kmsService.decrypt(
       encryptedAuthRequestJWE
     );
+
     const claims = await jwtService.getPayloadWithSigCheck(
       authRequestJweDecryptedAsJwt
     );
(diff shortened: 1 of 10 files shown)


src/components/authorize/jwt-service.ts
```diff
@@ -29,13 +30,55 @@ export class JwtService implements JwtServiceInterface {
...
+      "scope",
+      "redirect_uri",
+    ];
+    expectedkeys.forEach((claim) => {
+      if (!Object.prototype.hasOwnProperty.call(claims, claim)) {
+        throw new ClaimsError(`${claim} claim missing`);
+      }
+    });
+
+    if (claims.exp <= Math.floor(new Date().getTime() / 1000)) {
```
JHjava, line 74: In the AuthRequest it explicitly sets it to UTC. Do we need to explicit here and use UTC?

src/components/authorize/authorize-controller.ts
```diff
@@ -32,12 +38,27 @@ function createConsentCookie(
...
+      const authRequestJweDecryptedAsJwt = await kmsService.decrypt(
+        encryptedAuthRequestJWE
+      );
+
+      const claims = await jwtService.getPayloadWithSigCheck(
+        authRequestJweDecryptedAsJwt
+      );
+
+      jwtService.validateClaims(claims);
+    }
```
JHjava, line 60: We should probably throw when it is undefined

JHjava (commented):
I've left a couple of comments. 

In addition, we also need to validate the query parameters as well as the claims in the JWT. This will be based on the required values in the OAuth spec.

src/components/authorize/jwt-service.ts
gn-GDS: The date is passed across as unix time, and is validated using unix time. We shouldn't need worry about timezones?

JHjava: If that is the case, then great

src/components/authorize/authorize-controller.ts
gn-GDS: validateClaims always either returns claims or throws an error. This will be changed once we know what were doing with the claims.

src/components/authorize/jwt-service.ts
```diff
@@ -29,13 +31,65 @@ export class JwtService implements JwtServiceInterface {
...
+    const errors: string[] = [];
+    const requiredclaims = getKnownClaims();
+
+    Object.keys(requiredclaims).forEach((claim) => {
+      if (requiredclaims[claim] !== claims[claim]) {
+        errors.push(`${claim} has incorrect value`);
+      }
+    });
+
+    if (claims.exp <= Math.floor(new Date().getTime() / 1000)) {
```
mrwilson, line 83: We're doing manual checks for expired (`exp`) and not yet valid (`nbf`) here - does the underlying JOSE library we're using have the capability to do these checks for us?

isaac-GDS: We don't currently use the JOSE library in the frontend - the check here seems too small to bring in a library to do ?

src/components/authorize/authorize-controller.ts
isaac-GDS: if statement was unnecessary - removed so change in comment not needed.

src/components/authorize/jwt-service.ts
mrwilson: Not just for checking these but for validation of JWE/S in general - I'd much rather us use a library that has thought about lots of different edge cases than rolling our own validation, if we can.

https://github.com/panva/jose is heavily used in the rest of the programme and has built-ins for JWE/S.

I appreciate that the fact we use KMS complicates things, but it's still worth checking out if we can delegate our structural validation to something else.

isaac-GDS: Looked into using JOSE . As you mentioned there was complications with trying to use it in our decryption due to KMS but have moved to using JOSE for our validation.

src/components/authorize/claims-config.ts
```diff
@@ -0,0 +1,38 @@
+export const expectedClientId = "orchestrationAuth";
```
ayoshebby, line 1: For adherence to naming conventions, I would capitalize the immutable variable. Consider using `EXPECTED_CLIENT_ID` or `CLIENT_ID`.

src/components/authorize/tests/authorize-integration.test.ts
```diff
@@ -4,9 +4,17 @@ import { sinon } from "../../../../test/utils/test-utils";
...
+  AuthorizeServiceInterface,
+  JwtServiceInterface,
+  KmsDecryptionServiceInterface,
+  StartAuthResponse,
+} from "../types";
 import { createApiResponse } from "../../../utils/http";
 import { AxiosResponse } from "axios";
+import { createmockclaims, getPrivateKey, getPublicKey } from "./test-data";
+import { JwtService } from "../jwt-service";
+import { createJwt } from "./test-data";
```
DTaylorGDS, line 17: Sonar just wants you to roll this in with the import a couple of lines up - I guess you already saw, but just in case ^

src/components/authorize/claims-config.ts
```diff
@@ -0,0 +1,38 @@
...
+
+export type Claims = typeof claimsInstance;
+// construct object for type
+const claimsInstance = getClaimsObject();
+// eslint-disable-next-line @typescript-eslint/explicit-module-boundary-types
+export function getClaimsObject() {
+  return {
+    iss: "",
+    aud: "",
+    exp: NaN,
```
DTaylorGDS, line 22: Genuine question - what's the deal with NaN as opposed to null or undefined to indicate uninitialized values (Slack if easier and you have time to explain)

src/components/authorize/authorize-controller.ts
```diff
@@ -132,3 +157,17 @@ export function authorizeGet(
     return res.redirect(redirectPath);
   };
 }
+
+function validateQueryParams(clientId: string, responseType: string) {
+  if (clientId === undefined) {
```
DTaylorGDS, line 162: Do we need to check for null in these or is that impossible?

src/components/authorize/claims-config.ts
isaac-GDS: NaN give a number type which is needed when converting this object to a type.

DTaylorGDS: Cool - I guess that makes sense. Typescript :(

PR #1134, opened 2023-09-06
AUT-1648: validate the jwt claim object and set in the session

BeckaL:
## What?

Orchestration currently passes scopes and claims to authentication. We decided to simplify this, by only passing claims when the auth/orch split feature flag is on.  There, we need to parse this claims object and set it in the session, in order to forward the request on (this will be done in a subsequent PR).

This should not affect current behaviour, so we still parse scopes.

[PR](https://github.com/alphagov/di-authentication-api/pull/3313) to pass only claims through on the backend

### DIFF
--- a/src/components/authorize/authorize-controller.ts
+++ b/src/components/authorize/authorize-controller.ts
@@ -88,9 +88,15 @@ export function authorizeGet(
 
     req.session.client.prompt = loginPrompt;
 
+    if (claims.claim !== undefined) {
+      const claim = JSON.parse(claims.claim);
+      if (claim.userinfo !== undefined) {
+        req.session.client.claim = Object.keys(claim.userinfo);
+      }
+    }
+
     req.session.client.serviceType = claims.service_type;
     req.session.client.name = claims.client_name;
-    req.session.client.scopes = claims.scope.split(" ");
     req.session.client.cookieConsentEnabled = claims.cookie_consent_shared;
     req.session.client.redirectUri = claims.redirect_uri;
     req.session.client.state = claims.state;
--- a/src/components/authorize/claims-config.ts
+++ b/src/components/authorize/claims-config.ts
@@ -25,7 +25,6 @@ export type Claims = {
   confidence: string;
   state: string;
   client_id: string;
-  scope: string;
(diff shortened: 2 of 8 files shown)


src/components/authorize/claims-config.ts
```diff
@@ -9,30 +9,43 @@ export function getKnownClaims(): {
...
+  client_name: string;
+  cookie_consent_shared: boolean;
+  consent_required: boolean;
+  is_one_login_service: boolean;
+  service_type: string;
+  govuk_signin_journey_id: string;
+  confidence: string;
+  state: string;
+  client_id: string;
+  scope: string;
```
DTaylorGDS, line 28: Where is this scope object coming from? I might have misunderstood (probably) but I thought you were getting rid of it from the claims JWT?: https://github.com/alphagov/di-authentication-api/pull/3313/files#diff-e0d0ad45a3807a8326404be3bcf0d979b5fd01235b7a751394012ca6a4c708c7L387

src/components/authorize/authorize-controller.ts
```diff
@@ -89,6 +89,13 @@ export function authorizeGet(
...
     req.session.client.serviceType = claims.service_type;
+    if (claims.claim !== undefined) {
+      const claim = JSON.parse(claims.claim);
+      if (claim.userinfo !== undefined) {
+        req.session.client.claim = Object.keys(claim.userinfo);
+      }
+    }
+
     req.session.client.name = claims.client_name;
     req.session.client.scopes = claims.scope.split(" ");
```
DTaylorGDS, line 100: Basically a duplicate of the other question, but what is `claims.scope`?

src/components/authorize/claims-config.ts
BeckaL: Good point, will remove, we were thinking to keep it to avoid breaking current behaviour, but not realising the auth controller is not in current use

src/components/authorize/authorize-controller.ts
BeckaL: As above, good point, we can remove

DTaylorGDS (approved):

PR #1137, opened 2023-09-11
Aut 1646/call new authcode lambda

BeckaL:
## What?

We have a new Lambda for generating Auth codes for Authentication. This is the AuthenticationAuthCodeHandler which sits in di-authentication-api. 

In the Authentication frontend, we call the AuthCodeHandler from the auth-code-controller.ts . When the Auth Orch feature flag is enabled, we need to call the new AuthenticationAuthCodeHandler which lives on the endpoint /orch-auth-code on the Authentication internal API

When the Authentication frontend receives the response it will be able to redirect back to Authentication.

## Why?

This is part of enabling the split of authentication and orchestration.

### DIFF
--- a/src/components/auth-code/auth-code-service.ts
+++ b/src/components/auth-code/auth-code-service.ts
@@ -1,4 +1,4 @@
-import { ApiResponseResult } from "../../types";
+import { ApiResponseResult, UserSessionClient } from "../../types";
 import { API_ENDPOINTS } from "../../app.constants";
 import {
   createApiResponse,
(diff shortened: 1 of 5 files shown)


src/components/auth-code/auth-code-service.ts
```diff
@@ -7,25 +7,38 @@ import {
...
-      getRequestConfig({
-        baseURL: getApiBaseUrl(),
-        sessionId: sessionId,
-        clientSessionId: clientSessionId,
-        sourceIp: sourceIp,
-        persistentSessionId: persistentSessionId,
-      })
-    );
+    const config = getRequestConfig({
+      baseURL: getApiBaseUrl(),
```
DTaylorGDS, line 21: I thought that a difference between pre and post split is that the current Auth Code lambda is part of the OIDC API whereas the new one is part of the Frontend (Internal) API. So then the base URL would be different. Do you agree?

BeckaL: Ooooh I'd missed that as a nuance, but agreed.

src/components/auth-code/auth-code-service.ts
```diff
@@ -7,25 +7,38 @@ import {
...
+      sessionId: sessionId,
+      clientSessionId: clientSessionId,
+      sourceIp: sourceIp,
+      persistentSessionId: persistentSessionId,
+    });
+
+    let endpoint: string;
+
+    if (supportAuthOrchSplit()) {
+      config.data = {
```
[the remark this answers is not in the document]
BeckaL: I'm not sure if you have a specific thing that you're not sure about here so sorry if this is not what you're asking, but "data" is the axios request field for the body that you send with the request (docs here), so this is effectively just setting the body we need. Does that answer your question or am I missing something?

src/components/auth-code/auth-code-service.ts
BeckaL: Will [this environment variable](https://github.com/alphagov/di-authentication-frontend/blob/a903f81fac8efb4f00d1815ea10d35fee4798ca4/src/config.ts#L14) be pointing to the right place, or do we need a new one?

DTaylorGDS: That's the one, yeah 👍

BeckaL: Thanks, updated

src/components/auth-code/auth-code-service.ts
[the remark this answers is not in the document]
BeckaL: For history, me and Dom discussed and boiled down to needing to make a post request in the new world, have amended

DTaylorGDS (approved):

src/components/auth-code/auth-code-service.ts
```diff
@@ -7,25 +7,48 @@ import {
...
+      clientSessionId: clientSessionId,
+      sourceIp: sourceIp,
+      persistentSessionId: persistentSessionId,
+    });
+
+    let response: AxiosResponse;
+
+    if (supportAuthOrchSplit()) {
+      const body = {
+        claim: clientSession.claim,
```
DTaylorGDS, line 40: I think these will need to match the backend's request fields?:

https://github.com/alphagov/di-authentication-api/blob/main/frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/entity/AuthCodeRequest.java#L9

So:

- "redirect-uri"
- "state"
- "claims"

DTaylorGDS (changes requested):

src/components/auth-code/auth-code-service.ts
BeckaL: 🤦‍♀️ good spot

DTaylorGDS (approved):

PR #1158, opened 2023-09-26
AUT-1650: Support triage page

gtvj:
### Some background

All pages currently provide direct links to the Contact Forms (at `/contact-us`) via a "feedback" link in the phase banner and a "support" link in the footer of all page. There are also _indirect_ routes to the Contact Forms through Privacy notice, Terms & Conditions and Accessibility statement links (to `/support`) which present a form asking users if they are members of the public or working for a government service (when a user selects the public option on this page they are routed to the contact forms).

### What this PR does

1. Allows the URL for contact forms to be set through a new `URL_FOR_SUPPORT_LINKS` environment variable so links can be go to the new Triage Page (and this can be tested in lower environments)
2. Retains `/contact-us` as a default where the environment variable is not provided
3. Adds the current page URL as an encoded `fromURL` query parameter where `URL_FOR_SUPPORT_LINKS` is set to the URL for the Contact Centre triage page (`fromURL` is required by the Triage Page - more detail is available in the Jira ticket)
4. Changes the `supportPost` method in `footer-page-controller.ts` to route public users to the `URL_FOR_SUPPORT_LINKS` URL

## Why?

As part of the Contact Centre project, a new triage page is being built and all support links from Auth are to go there.

### DIFF
--- a/src/locales/cy/translation.json
+++ b/src/locales/cy/translation.json
@@ -10,7 +10,11 @@
     "sendMessage": "Anfon neges",
     "phaseBanner": {
       "tag": "beta",
-      "content": "Mae hwn yn wasanaeth newydd – bydd eich <a  href=\"/contact-us?supportType=PUBLIC\" class=\"govuk-link\" rel=\"noopener\" target=\"_blank\">adborth (agor mewn tab newydd)</a> yn ein helpu i’w wella."
+      "message": {
+        "start": "Mae hwn yn wasanaeth newydd – bydd eich ",
+        "linkText": "adborth (agor mewn tab newydd)",
+        "end": " yn ein helpu i’w wella."
+      }
     },
     "authenticatorAppIssuer": "GOV.UK One Login",
     "yes": "Gallwch",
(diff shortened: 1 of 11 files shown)


src/locales/cy/translation.json
```diff
@@ -44,7 +48,7 @@
       "label": "Parhau"
     },
     "contactAccounts": {
-      "contactLinkHref": "https://signin.account.gov.uk/contact-us",
+      "contactLinkHref": "https://signin.account.gov.uk/support",
```
JHjava, line 51: I couldn't find usage for these anywhere in the code base. Are they still needed?

src/locales/en/translation.json
```diff
@@ -44,7 +48,7 @@
       "label": "Continue"
     },
     "contactAccounts": {
-      "contactLinkHref": "https://signin.account.gov.uk/contact-us",
+      "contactLinkHref": "https://signin.account.gov.uk/support",
```
JHjava, line 51: I couldn't find usage for these anywhere in the code base. Are they still needed?

JHjava (commented):
We don't have a common footer used across all the components in DI such as in IPV. Are any changes needed to these other footers?

src/locales/cy/translation.json
gtvj: Thanks James - I've just pushed a commit to removes these 👍🏽

JHjava (approved):

PR #1160, opened 2023-09-29
AUT-1736: Update Contact Form to consume Triage Page variables

gtvj:
## What?

Allows Contact Form application to:

1. handle GET parameters that will be included in links from the Triage Page
2. persist these through the journey
3. include the values in submissions to SmartAgent. 

The properties are: 

* `fromURL` which will be included in all links from the Triage Page. This holds the page URL that the user had been on before they went to the Triage Page. 
* Three additional parameters that will be included for ID Check App links coming from the Triage Page. These properties are `theme=id_check_app`, `appSessionId` and `appErrorCode`.

### Approach 

There are two parts to this: 

1. I've chosen to introduce a new route (`/contact-us-from-triage-page`) and controller method to handle requests coming from the Triage Page because links that include the `theme=id_check_app` query parameter will need to be redirected. This seemed better than adding branching logic to the `contactUsGet` handler for `/contact-us`)
6. Where these GET parameters are included, they persist through the Contact Forms through a combination of parameters passed to `res.redirect` and as hidden form fields. This follows the same approach that's used for the existing query parameters

## Why?

A Triage Page is being developed by the Home team that will send through a new `fromURL` query parameter in addition to existing parameters required by the ID Check App.

### DIFF
--- a/src/components/contact-us/contact-us-controller.ts
+++ b/src/components/contact-us/contact-us-controller.ts
@@ -82,6 +82,12 @@ export function contactUsGet(req: Request, res: Response): void {
   const REFERER = "referer";
 
   let referer = validateReferer(req.get(REFERER));
+  let fromURL;
+
+  if (req.query.fromURL) {
+    fromURL = validateReferer(req.query.fromURL as string);
+    logger.info(`fromURL query param received with value ${fromURL}`);
+  }
 
   if (req.query.referer) {
     referer = validateReferer(req.query.referer as string);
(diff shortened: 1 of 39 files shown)


src/components/contact-us/contact-us-controller.ts
```diff
@@ -111,6 +118,30 @@ export function contactUsGet(req: Request, res: Response): void {
   return res.render("contact-us/index-public-contact-us.njk", options);
 }
 
+export function contactUsGetFromTriagePage(req: Request, res: Response): void {
+  const queryParams = new URLSearchParams({
+    ...(validateAppId(req.query.appSessionId as string) && {
```
dbes-gds, line 123: what do the '...' do here?

gtvj: Thanks @dbes-gds - it uses [spread syntax](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax#conditionally_adding_properties_to_an_object) to conditionally add a property to the object literal that's passed to `URLSearchParams` where left hand side of the `&&` is truthy.

```typescript
  const queryParams = new URLSearchParams({
    ...(validateAppId(req.query.appSessionId as string) && {
      appSessionId: getAppSessionId(req.query.appSessionId as string),
    }),
    ...(getAppErrorCode(req.query.appErrorCode as string) && {
      appErrorCode: getAppErrorCode(req.query.appErrorCode as string),
    }),
    ...(validateReferer(req.query.fromURL as string) && {
      fromURL: validateReferer(req.query.fromURL as string),
    }),
  });
```

My main motivation was to avoid empty values appearing as search params in the query string - but this might not be the best way to code it (I went for it because it was more concise). I think the same could be achieved with code along these lines, but it seemed quite long:

```typescript
  const params = {};
  
  if(validateAppId(req.query.appSessionId as string)) {
    params.appSessionId = getAppSessionId(req.query.appSessionId as string)
  }

  if(getAppErrorCode(req.query.appErrorCode as string)) {
    params.appSessionId = getAppErrorCode(req.query.appErrorCode as string)
  }

  if(validateReferer(req.query.fromURL as string)) {
    params.fromURL = validateReferer(req.query.fromURL as string)
  }

  const queryParams = new URLSearchParams(params);
```

I'd be very happy to change the implementation to be along these lines though (or another way) 👍🏽

dbes-gds: @gtvj fine with this if it does the trick, just wasn't used to the syntax

dbes-gds (approved):

PR #1170, opened 2023-10-13
AUT-1650: Fix back links from Triage Page

gtvj:
## What?

Changes the "Back" link `href` on `/contact-us-further-information` from being a hard-coded value to either: 

1. the `referer` where the referer passes the checks in `validateReferer`, or
2. `/contact-us` where it does not

## Why?

When testing inbound links from the triage page that was deployed to Integration today, we found the links that include the `theme=id_check_app` parameter are redirected to `/contact-us-further-information` as expected but always use `/contact-us` as the back link `href`  (see the description in #1160 of how the presence of `theme=id_check_app` results in the user being passed directly to `/contact-us-further-information). This is problematic for two reasons: 

1. The user had not come from `/contact-us`
2. It resulted in the user getting stuck in a loop between `/contact-us` and `/contact-us-further-information` when clicking the back link

This was happening because the Back link `href` was hard-coded into the template. This had not been a problem previously because the only route to this page was via `/contact-us` but, with the introduction of the triage page, users have an alternative route to reach the page

## Related PRs

* #1160 Updated the contact forms to receive parameters from the triage page and introduced the routing.

### DIFF
--- a/src/components/contact-us/contact-us-controller.ts
+++ b/src/components/contact-us/contact-us-controller.ts
@@ -252,9 +252,13 @@ export function furtherInformationGet(req: Request, res: Response): void {
     return res.redirect(PATH_NAMES.CONTACT_US);
   }
 
+  const backLinkHref =
+    validateReferer(req.get("referer")) || PATH_NAMES.CONTACT_US;
+
   if (isAppJourney(req.query.appSessionId as string)) {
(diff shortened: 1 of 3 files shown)


gtvj:
Deleting this branch because work is being moved under a separate Jira ticket.

PR #1175, opened 2023-10-17
AUT-1650: Set environment variables for new Contact Centre

gtvj:
## What?

Sets the following environment variables consistently across all environments. After this change, all `.tfvars` files should reflect the values shown in the table below.

|  | Support SmartAgent  |  Support Welsh in Contact Forms | URL for support links  |
|---|---|---|---|
| Sandpit  | 1  | 1  | `https://home.build.account.gov.uk/contact-gov-uk-one-login`  |
| Build  | 1  | 1  | `https://home.build.account.gov.uk/contact-gov-uk-one-login`  |
| Integration  | 1  | 1  | `https://home.integration.account.gov.uk/contact-gov-uk-one-login`  |
| Staging  | 1  | 1  | `https://home.staging.account.gov.uk/contact-gov-uk-one-login`  |
| Live | 0  | 0  | `/contact-us`  |

## Why?

Prepares the environments for go-live later this month. After this PR, the only change necessary to release to production will be changing these environments to the values in `production.tfvars`

### DIFF
--- a/ci/terraform/production.tfvars
+++ b/ci/terraform/production.tfvars
@@ -12,6 +12,8 @@ support_international_numbers                                       = "1"
 support_account_recovery                                            = "1"
 support_smart_agent                                                 = "0"
 client_name_that_directs_all_contact_form_submissions_to_smartagent = "di-auth-stub-relying-party-production"
+support_welsh_language_in_support_forms                             = "0"
+url_for_support_links                                               = "/contact-us"
 
 logging_endpoint_arns = [
   "arn:aws:logs:eu-west-2:885513274347:destination:csls_cw_logs_destination_prodpython"
--- a/ci/terraform/build.tfvars
+++ b/ci/terraform/build.tfvars
@@ -16,6 +16,7 @@ account_recovery_code_entered_wrong_blocked_minutes                 = "0.5"
 code_request_blocked_minutes                                        = "0.5"
 code_entered_wrong_blocked_minutes                                  = "0.5"
 client_name_that_directs_all_contact_form_submissions_to_smartagent = ""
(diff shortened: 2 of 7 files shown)


ci/terraform/production.tfvars
```diff
@@ -12,6 +12,8 @@ support_international_numbers                                       = "1"
 support_account_recovery                                            = "1"
 support_smart_agent                                                 = "0"
 client_name_that_directs_all_contact_form_submissions_to_smartagent = "di-auth-stub-relying-party-production"
+support_welsh_language_in_support_forms                             = "0"
+url_for_support_links                                               = "/contact-us"
```
BeckaL, line 16: Is there a reason this is a path and the others are urls? Should this be the full url?

BeckaL (commented):
Looks generally good to me, just a question in there about a url / path

ci/terraform/production.tfvars
gtvj: Thanks @BeckaL - there is an explanation but I'm open to changing the approach. 

I'd opted to include the `url_for_support_links` variable in production so that all the environment variable keys are in place across all environments and to reduce the (admittedly very small) potential for the key to be misspelled before we go live. This is partly because I've been finding all the cognitive load of all the environment variables and AWS Secrets a bit of a challenge but also because I'm currently working on some "Go live" documentation and was thinking it's easier to say "Find and change" rather than "Add".

It should be redundant because omitting it _should_ result in the `"/contact-us"` path kicking in through the defaults. 

I'm very happy to take your view on it (this is definitely "backend" rather than "frontend" so I'm not best placed to make a call. Just let me know 👍🏽

DTaylorGDS (approved):

PR #1183, opened 2023-10-20
Add Dynatrace to the ECS container

melvabout:
## What?

Add the dynatrace container to the frontend.

## Why?

So that application metrics will be ingested into dynatrace.

- [ ]


dbes-gds:
Could we do a switch to rollout per environment?

melvabout:
@dbes-gds, @two-first-names  You could add conditionals into the docker file, but that wouldn't give you deployment  control per environment. I can't see a terraform apply in the workflows and it looks like a zip of terraform is created, so I assume that secure pipeline calls the terraform apply command? I think the only way to control this then would be to have two Docker images one with dynatrace and one with out and then you could use the variables `image_uri`, `image_tag` and `image_digest` in each environment to control the deployment. I don't have access to the aws accounts so this is an assumption. Thoughts?

BeckaL:
Closed following conversation with @two-first-names - this is old and superseded

PR #1202, opened 2023-11-03
AUT-1624: Test standard ECS auto scaling configuration in sandpit, build and staging

dbes-gds:
## What?

Test standard ECS auto scaling configuration in sandpit, build and staging.

Switch off the current auto scaling configuration and switch on the standard one in these environments. Have applied the standard configuration using Terraform: 

https://github.com/govuk-one-login/architecture/blob/4b985350c63b000343bccc76027101139b0e5e35/adr/01XX-fargate-scaling.md

## Why?

The current configuration did not scale appropriately during performance test.


whi-tw (approved):
All looks good.

Need to make sure we fully replace the vestigial 'v1' code with the new 'v2' after performance testing passes, as-per the TODO!

PR #1214, opened 2023-11-09
AUT-1497: Add "What service were you trying to use?" to the required …

LazarAlexandru-Constantin:
## What?

Update user supports forms to add 'What service were you trying to use?' field.
This ticket is to add a ‘What service were you trying to use?’ field and hint text to a number of user support forms.
In all cases, it should be placed just above the warning component.

## Why?

As One Login User Support
I want to know what service a user was trying use
So that I can resolve their problem quicker

### DIFF
--- a/src/components/contact-us/contact-us-controller.ts
+++ b/src/components/contact-us/contact-us-controller.ts
@@ -502,6 +502,10 @@ export function getQuestionsFromFormTypeForMessageBody(
         "pages.contactUsQuestions.accountNotFound.section2.header",
         { lng: "en" }
       ),
+      serviceTryingToUse: req.t(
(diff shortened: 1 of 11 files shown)


gtvj:
Hi @LazarAlexandru-Constantin, I've been testing these forms and have noticed an issue with the "Your GOV.UK One Login cannot be found" in that we're asking the user the same question twice (as shown in the screenshot below). I'm pretty sure this isn't what UCD colleagues would want to happen. Could you check with them and make any necessary changes? 

Screenshot 2023-11-14 at 12 10 28

gtvj (changes requested):
Hi Alex, I'm afraid I've found that at least one form has the "What service were you trying to use?" question duplicated.

gtvj (approved):

PR #1217, opened 2023-11-13
ATO:195 Amend WAF query param limit

ethanmills:
## What?
Amend WAF query param limit

## Why?
The authorize request from orchestration exceeds the limit imposed by this WAF rule. Maintain for all other endpoints, but extend to 4096 for the authorize endpoint.

## Related PRs
https://github.com/govuk-one-login/authentication-api/pull/3563

### DIFF
--- a/ci/terraform/waf.tf
+++ b/ci/terraform/waf.tf
@@ -43,6 +43,9 @@ resource "aws_wafv2_web_acl" "frontend_alb_waf_regional_web_acl" {
         excluded_rule {
           name = "GenericRFI_BODY"
         }
+        excluded_rule {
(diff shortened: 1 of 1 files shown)


CarlyG55 (dismissed):

sebibo:
Primary concern is the increased likelihood of a denial of service impact of the proposed change - global changes to the WAF for accounts would affect all the APIs not just the specific API which has the issue. 

From a security perspective the recommendation would one of the following presented in preferred order:
1. reduce the query size to be under 2000 characters
2. create a specific rule for the single API call affected 
3. increase the rate limit for all API calls - but by the smallest amount possible

billrdunn (approved):
LGTM

PR #1218, opened 2023-11-14
AUT-1643: Create capability to run acceptance tests against a local RP/front-end

ayoshebby:
## What?

1. Replaced multiple mini stubs in Docker Compose with a single, unified stub.
2. Implemented an Express-based solution to serve HTML, replicating functionality from existing deployed stubs (e.g., Staging stub)
3. Introduced a new local endpoint in a standalone micro stub (using Express), replacing the logic in dev-app.js.
4. This endpoint is now parameterized based on form selection in the UI, similar to what we have for example (without styling) [here](https://di-auth-stub-relying-party-staging.london.cloudapps.digital/) 
5. Ensured the new endpoint returns a 302 response to localhost:3000, mirroring the behavior of existing local stubs, including cookie setting functionality.

## Why?

Allow acceptance tests to run against a local front-end


micro-stub/micro-stub-server.js
```diff
@@ -0,0 +1,125 @@
...
+const path = require("path");
+
+app.get("/", (req, res) => {
+  res.sendFile(path.join(__dirname, 'index.html'))
+})
+
+app.get("/authorize", (req, res) => {
+  const authRequest = getAuthorizeRequest(req.query);
+
+  console.log("Query parameters passed from form:");
```
dbes-gds, line 19: Some console.log failing the lint check.

PR #1222, opened 2023-11-17
AUT-1497: Fix [undefined] service field for contact support forms 

LazarAlexandru-Constantin:
## What?

This populates the service field in the support tickets with the correct value for the "something else" and "technical error" support pages. 

## Why?

After merging the previous branch for the same user story, it was noticed that the requests sent from the "something else" and "technical error" had the value of service field as [undefined].

These changes have been verified locally, by checking if the value of the required field changed from [undefined] to the correct value.

## Related PRs

https://github.com/govuk-one-login/authentication-frontend/pull/1214


BeckaL (approved):
LGTM

PR #1225, opened 2023-11-22
Aut 1783/refactor asset js

mark-cab:
## What?

Modular refactor of the assets javascript code and splitting analytics code to make security review easier.

## Why?

Pre-requisite for GA4 and want to add tests to the code as none are present currently.


gtvj:
Hi @mark-cab, I've just pulled the branch and run the client-side build task, but when I visit the home page in 'fresh' browser (i.e. one with no history), I'm not shown the cookie banner and the value of `cookie_preferences_set` is `analytics: true`. I'd have expected to be shown the cookie banner until I've set my preferences and for the value of `cookie_preferences_set` to reflect my choice. Apologies in advance if this is happening because I'm missing something, but the page is loading the new scripts.

Would you mind checking with @CraigRyder if we have adequate acceptance tests to ensure the cookie setting functionality is behaving as expected? I'd ideally like to see tests that verify cookies are set appropriately because it seems there's significant potential for third party cookies being set due to a bug in the client-side JavaScript.

gtvj (changes requested):
Hi @mark-cab, I've left quite a few comments and suggestions. I hope they all sound OK. Let me know if you'd like me to chat them through or if there's anything I can help with.

PR #1227, opened 2023-11-24
AUT-2029: Turn on frame ancestors switch in build and integration

LazarAlexandru-Constantin:
## What?

Turn on frame ancestors switch in build and integration.

## Why?

We need to fix the identified issues with the Content Security Policy (CSP) frame ancestors headers.


ci/terraform/integration.tfvars
```diff
@@ -16,3 +16,5 @@ url_for_support_links                                               = "https://h
 logging_endpoint_arns = [
   "arn:aws:logs:eu-west-2:885513274347:destination:csls_cw_logs_destination_prodpython"
 ]
+
```
dbes-gds, line 19: let's do staging rather than integration

dbes-gds (approved):

PR #1231, opened 2023-11-29
AUT-1860: Added Account Interventions Feature Flag

alhcomer:
## What?

Added feature flag for the account interventions Lambda.

## Why?

Feature flag has been added to control whether account interventions is turned on or not. This is so the account interventions can be tested on an environment-by-environment basis.


DTaylorGDS (approved):
LGTM - as discussed, the environment overrides aren't really doing anything since they match the default value, but it's not a problem to keep them if we think that's more explicit in some way

PR #1235, opened 2023-11-30
BAU: Deploy frontend to sandpit using AWS SSO rather than GDS CLI

dbes-gds:
## What?

Deploy frontend to sandpit using AWS SSO rather than GDS CLI

## How to configure

Relies on the creation of two local AWS profiles with standard names:

auth-dev-tools-admin = digital-identity-tools-dev (role = di-tools-dev-admin)
auth-dev-admin = gds-di-development (role = authentication-admin)

Create the profiles using:

aws configure sso

SSO session name (Recommended): [enter a session name]
SSO start URL [None]:[enter the url for the AWS SSO console]
SSO region [None]: eu-west-2
SSO registration scopes [sso:account:access]: [no value needed]

--> Attempting to automatically open the SSO authorization page in your default browser.
-->If the browser does not open or you wish to use a different device to authorize this request, open the following URL:

--> https://device.sso.eu-west-2.amazonaws.com/

-->Then enter the code:

There are 209 AWS accounts available to you. [choose the correct account for the profile, see above]
Using the account ID xxxxxxxxx
[choose the correct role for the profile, see above]
Using the role name "xxxxxxxxxxx"
CLI default client Region [None]: eu-west-2
CLI default output format [None]: [none required]
CLI profile name [di-tools-dev-admin-706615647326]: [enter the standard role name, see above]

To use this profile, specify the profile name using --profile, as shown:

aws s3 ls --profile auth-dev-tools-admin

Check that the profile is working by running the command suggested.

## Why?

Not all devs have access to the GDS CLI, and we are trying to phase it out.
Script needed rewriting as it relies on two AWS accounts in a single execution.

### DIFF
--- a/deploy-sandpit.sh
+++ b/deploy-sandpit.sh
@@ -55,11 +55,12 @@ while [[ $# -gt 0 ]]; do
 done
 
 echo "Generating temporary ECR credentials..."
-gds aws di-tools-dev -- aws ecr get-login-password --region eu-west-2 | docker login --username AWS --password-stdin "${REPO_URL}"
+eval "$(aws configure export-credentials --profile auth-dev-tools-admin --format env)"
+aws ecr get-login-password --region eu-west-2 | docker login --username AWS --password-stdin "${REPO_URL}"
 
 if [[ $BUILD == "1" ]]; then
   echo "Building image..."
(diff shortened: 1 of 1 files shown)


BeckaL (approved):
Have confirmed this works for me

PR #1236, opened 2023-12-01
encode form input data to prevent query params content injection pass…

mark-cab:
…through

## What?

To prevent content injection through the hidden fields used within contact us we had to encode the value being passed to the frontend so malicious scripts were not run on clients computers.

## Why?

Raised in a pen test, [issue](https://govukverify.atlassian.net/browse/AUT-1953) was created as a result.
Attempts were first made to replace and encode at different points but the safest method was encoding during get and decoding during the post.


gtvj (commented):
Hi @mark-cab - just one small comment about function naming. I did also wonder if the commit history might be condensed a bit, but I'll leave that to you to decide.

gtvj (dismissed):

gtvj (approved):

PR #1240, opened 2023-12-04
AUT-1862: Created Page and Route for Temporary Suspended Page

alhcomer:
## What?

Added template, route, unit test and translations for the Temporary Suspended page.

## Why?

The changes are necessary to be implemented as part of the account interventions user journey.

| English Translation | Welsh Translation | 
| --------------- | --------------- | 
| English Translation  | welsh-temporary-blocked-screenshot  |

### DIFF
--- a/src/app.constants.ts
+++ b/src/app.constants.ts
@@ -69,6 +69,7 @@ export const PATH_NAMES = {
   CHANGE_SECURITY_CODES_CONFIRMATION: "/change-codes-confirmed",
   PASSWORD_RESET_REQUIRED: "/password-reset-required",
   UNAVAILABLE_PERMANENT: "/unavailable-permanent",
+  UNAVAILABLE_TEMPORARY: "/unavailable-temporary",
 };
 
 export const HREF_BACK = {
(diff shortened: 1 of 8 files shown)


src/app.constants.ts
```diff
@@ -69,6 +69,7 @@ export const PATH_NAMES = {
   CHANGE_SECURITY_CODES_CONFIRMATION: "/change-codes-confirmed",
   PASSWORD_RESET_REQUIRED: "/password-reset-required",
   UNAVAILABLE_PERMANENT: "/unavailable-permanent",
+  UNAVAILABLE: "/unavailable",
```
BeckaL, line 72: A check here - should this be `unavailable-temporary`? I see the ticket is confused on this (says both `unavailable` and `unavailable-temporary` but the latter seems to follow the pattern of the above

BeckaL (commented):
Generally looks sensible to me, nice one :-) Just a question here about the URL. Also, if you could put screenshots in the PR description that'd be great - we try to do this for any changes to the frontend. Either a before and after if we're changing something that exists already, or just the new pages in this case, in English and Welsh.

src/app.constants.ts
alhcomer: Ahh yes. Thanks for the reminder!

BeckaL (approved):
Nice, looks good to me

PR #1242, opened 2023-12-05
ATO-224: Return to orchestration on ID journeys

ethanmills:
## What?

Invoke the auth code logic in provide identity when the auth-orch split is enabled

## Why?

When the auth orch split is enabled, we should return to orchestration rather than redirecting to IPV.


ethanmills:
I will disable the prod pipeline and test this in staging with the split turned off before merging

dbes-gds (approved):

PR #1262, opened 2023-12-20
AUT-2116: Integrate 'Enter password' into the frontend re-authentication flow

ayoshebby:
## What?

Temporarily redirect the user to /enter-password during the re-authenticate journey only when feature flag is on
- Update state machine
- Update enter-password GET controller method
- Call new CheckReauthUsers API endpoint

## Why?

Temporarily redirect to /enter-password page whilst discussions are ongoing with HMRC regarding the new reauth screen

### DIFF
--- a/src/components/common/state-machine/state-machine.ts
+++ b/src/components/common/state-machine/state-machine.ts
@@ -127,7 +127,7 @@ const authStateMachine = createMachine(
               cond: "isConsentRequired",
             },
             {
-              target: [PATH_NAMES.SIGN_IN_OR_CREATE],
+              target: [PATH_NAMES.ENTER_PASSWORD],
               cond: "isReauthenticationRequired",
             },
             { target: [PATH_NAMES.AUTH_CODE], cond: "isAuthenticated" },
--- a/src/components/enter-password/enter-password-controller.ts
+++ b/src/components/enter-password/enter-password-controller.ts
@@ -19,6 +19,9 @@ import { MFA_METHOD_TYPE } from "../../app.constants";
 import xss from "xss";
 import { EnterEmailServiceInterface } from "../enter-email/types";
 import { enterEmailService } from "../enter-email/enter-email-service";
+import { supportReauthentication } from "../../config";
+import { CheckReauthServiceInterface } from "../check-reauth-users/types";
+import { checkReauthUsersService } from "../check-reauth-users/check-reauth-users-service";
 
 const ENTER_PASSWORD_TEMPLATE = "enter-password/index.njk";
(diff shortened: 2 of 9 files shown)


src/components/enter-password/enter-password-controller.ts
```diff
@@ -29,8 +32,36 @@ const ENTER_PASSWORD_ACCOUNT_EXISTS_TEMPLATE =
 const ENTER_PASSWORD_ACCOUNT_EXISTS_VALIDATION_KEY =
   "pages.enterPasswordAccountExists.password.validationError.incorrectPassword";
 
-export function enterPasswordGet(req: Request, res: Response): void {
-  res.render(ENTER_PASSWORD_TEMPLATE);
+export function enterPasswordGet(
+  service: CheckReauthServiceInterface = checkReauthUsersService()
+): ExpressRouteFunc {
+  return async function (req: Request, res: Response) {
+    if (!supportReauthentication()) {
```
BeckaL, line 39: I'm wondering if for readability we also pull the `isReauthenticationRequired` condition up to this bit. So the condition here is if either we don't support reauth, or reauth is not required, then early return and none of the rest of this code is relevant to the code path

src/components/common/state-machine/state-machine.ts
```diff
@@ -131,6 +131,10 @@ const authStateMachine = createMachine(
               cond: "isReauthenticationRequired",
             },
             { target: [PATH_NAMES.AUTH_CODE], cond: "isAuthenticated" },
+            {
+              target: [PATH_NAMES.ENTER_PASSWORD],
```
BeckaL, line 135: This condition will have no effect since the condition two above this will take precedence. (Did you write a test to check this did what you expected it to? This would be a good thing to TDD) I think instead we should be changing the `TARGET` in the condition two above this to go to `ENTER_PASSWORD` instead of `SIGN_IN_OR_CREATE`. There are tests already, this would be a good thing to TDD.

src/components/common/state-machine/state-machine.ts
```diff
@@ -146,6 +150,10 @@ const authStateMachine = createMachine(
               cond: "isIdentityRequired",
             },
             { target: [PATH_NAMES.SIGN_IN_OR_CREATE] },
+            {
```
BeckaL, line 153: I don't think we want this condition - if there's no existing session, they should just be going through the normal sign in or create path I think. There are also no tests for this case.

BeckaL (commented):
Haven't reviewed everything in detail but a couple of things that I think need changing here

BeckaL (approved):
Looks good to me

PR #1278, opened 2024-01-10
AUT-1958: Remove rate limit for GDS IPs in Staging

DTaylorGDS:
- This will allow performance testers to bypass normal rate limits on App Load Balancer
- In turn this will allow them to exercise the infrastructure in a way similar to a distibuted attack

## What?

Please include a summary of the change.

## Why?

Please include reason for the change and any other relevant context.

- [ ] Changes to the user interface have been demonstrated

Delete this section if the PR does not change the UI.

## Related PRs

Please include links to PRs in other repositories relevant to this PR.
Delete this section if not needed.


ci/terraform/waf.tf
```diff
@@ -1,3 +1,22 @@
...
+  ip_address_version = "IPV4"
+
+  addresses = [
+    "217.196.229.77/32",
+    "217.196.229.79/32",
+    "217.196.229.80/32", # (BYOD VPN Only)
+    "217.196.229.81/32",
+    "51.149.8.0/25",   # (GDS and CO VPN)
+    "51.149.8.128/29", # (BYOD VPN only)
+  ]
```
pskushwaha1, line 15: I would Suggest to add old IPs as well here , staging stubs are still using  OLD IPs

ci/terraform/waf.tf
```diff
@@ -169,6 +199,25 @@ resource "aws_wafv2_web_acl" "frontend_alb_waf_regional_web_acl" {
     }
   }
 
+  rule {
+    action {
+      block {}
+    }
+    priority = 60
```
pskushwaha1, line 206: This will lower the priority for Other Env also ? Since  AWS WAF evaluates  from the lowest numeric priority setting 

May be logic to set highest priority if environment is staging ?  leaving the priority untouched for other Env

ci/terraform/waf.tf
DTaylorGDS: Thanks @pskushwaha1 - I've added them just in case

ci/terraform/waf.tf
DTaylorGDS: Yes, fair point. Changed, thanks

pskushwaha1 (approved):

PR #1285, opened 2024-01-12
AUT-1997: Setup Password Reset Enter Authenticator App Code flow

LazarAlexandru-Constantin:
## What?

Change the screen-flow to include the Enter Authenticator App Code 2FA step in password reset when the switch is on.

## Why?

The BAU flow for the password reset is: 1.Check-email 2.Reset-Password 3.MFA. The order needs to be changed as mentioned above.

### DIFF
--- a/src/components/common/state-machine/state-machine.ts
+++ b/src/components/common/state-machine/state-machine.ts
@@ -507,7 +507,11 @@ const authStateMachine = createMachine(
           [USER_JOURNEY_EVENTS.RESET_PASSWORD_CODE_VERIFIED]: [
             {
               target: [PATH_NAMES.RESET_PASSWORD_2FA_SMS],
-              cond: "support2FABeforePasswordReset",
+              cond: "requiresResetPasswordMFASmsCode",
+            },
+            {
+              target: [PATH_NAMES.RESET_PASSWORD_2FA_AUTH_APP],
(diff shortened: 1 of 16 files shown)


src/components/common/state-machine/state-machine.ts
```diff
@@ -738,6 +747,15 @@ const authStateMachine = createMachine(
       requiresMFAAuthAppCode: (context) =>
         context.mfaMethodType === MFA_METHOD_TYPE.AUTH_APP &&
         context.requiresTwoFactorAuth === true,
+      requiresMFASmsCode: (context) =>
```
mark-cab, line 750: forget to clean this up?

LazarAlexandru-Constantin: No, I left this up on purpose, the requiresMFAAuthAppCode is a gate that's used for the login journey, I created 2 new gates for the Reset Password flow, that ensure that the user has mfa set up, that it is a password reset journey and that the switch is turned on. This was the approach I came up with to be entirely sure that the code is clear and that it works well.  I tried other approaches but for them there were integration tests and unit tests that were failing. I am open to suggestions for any better approach though.

mark-cab (approved):

mark-cab (approved):

PR #1290, opened 2024-01-17
AUT-2207: Send user journey to backend API when requesting or submitting MFA codes during reauthentication

mattvot:
## What?

Send the user journey to the backend API when they request or submit an MFA code during a reauthentication journey. Also took the opportunity to centralise the logic that determines the journey from the user session.

## Why?

We need to increment a relevant counter on the backend when the user requests and submits their MFA codes during reauthentication.


src/components/enter-password/tests/enter-password-controller.test.ts
```diff
@@ -47,6 +47,96 @@ describe("enter password controller", () => {
...
+        };
+        req.body["password"] = "password";
+
+        await enterPasswordPost(
+          false,
+          fakeService,
+          fakeMfaService
+        )(req as Request, res as Response);
+
+        expect(fakeMfaService.sendMfaCode).to.have.been.called;
```
BeckaL, line 87: Is it worth adding in the explicit expectations for what this is called with i.e. with an undefined journey type? (Or a separate test that is the negation of the below test, so "should not send a journey type to the MFA service when the user session does not have the reauthenticate property"

src/components/enter-password/tests/enter-password-controller.test.ts
```diff
@@ -193,13 +193,13 @@ describe("enter password controller", () => {
...
 
         expect(fakeMfaService.sendMfaCode).to.have.been.calledWith(
-          res.locals.sessionId,
-          res.locals.clientSessionId,
-          req.session.user.email,
-          "127.0.0.1",
-          res.locals.persistentSessionId,
-          false,
-          "",
+          sinon.match.string,
```
BeckaL: Like this, makes it much clearer the thing we actually care about here :-)

BeckaL (commented):
Nice one, generally looks good to me, one thought about cleaning up a test but non-blocking.

Before I approve, could you reword each commit to append the ticket number? This is our convention, so your commits should be "AUT-2207: Rest of commit message first line here"

src/components/enter-password/tests/enter-password-controller.test.ts
```diff
@@ -47,6 +47,96 @@ describe("enter password controller", () => {
...
+          false,
+          fakeService,
+          fakeMfaService
+        )(req as Request, res as Response);
+
+        expect(fakeMfaService.sendMfaCode).to.have.been.called;
+      });
+
+      it("should send REAUTHENTICATION journeyType to MFA service when user session has reauthenticate property", async () => {
+        const fakeService: EnterPasswordServiceInterface = {
```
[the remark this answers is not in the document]
mattvot: Some of these test setups could be brought up to the `beforeEach` calls I think. I'd vote we leave for now in this PR. We could address a general tidy up in a later PR.

BeckaL (approved):
LGTM! Thanks :-)

PR #1292, opened 2024-01-17
Set up ‘Check your phone’ and 'Auth app' flows for Password Reset Required

mark-cab:
## What?

Due to the way we have completed the MFA password update journey and the way forced password reset routing works, only the password update controller required updating to handle a forced password reset journey.

## Why?

We needed to add MFA to the forced password reset journey caused by the user using a weak password during login.
At some point we used routes to route the user to existing password reset functionality despite there being a state machine flow for the forced password reset journey, this made the MFA journey change a lot more simple once the difference from normal flow was understood.

related links:
[SMS journey](https://govukverify.atlassian.net/browse/AUT-2032)
[Auth App journey](https://govukverify.atlassian.net/browse/AUT-2033)


mark-cab:
closed as https://github.com/govuk-one-login/authentication-frontend/pull/1293 opened

PR #1304, opened 2024-01-22
AUT-2128: Pass in ID Token when reauthenticating

ayoshebby:
## What?

Add `id_token_hint` to the local stub

- Pass in the id token via a new environment variable `ID_TOKEN_VALUE`

## Why?

So we can test the re-authentication flow

### DIFF
--- a/dev-app.js
+++ b/dev-app.js
@@ -22,6 +22,7 @@ function createAuthorizeRequest() {
     "&client_id=" + process.env.TEST_CLIENT_ID +
     "&cookie_consent=accept" +
     "&_ga=test" +
+    getIDToken() +
     getUILocales();
 }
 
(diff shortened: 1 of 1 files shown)


dev-app.js
```diff
@@ -10,6 +10,7 @@ const port = process.env.PORT;
 
 function createAuthorizeRequest() {
   const vtr = process.env.VTR ? "vtr=" + encodeURI(process.env.VTR) : "";
+  const idToken = process.env.SUPPORT_REAUTHENTICATION ? "&id_token_hint=" + process.env.ID_TOKEN_VALUE : ""
```
BeckaL, line 13: I'm not sure if we want this always set when the feature flag is turned on, as it'll make every journey a re-auth journey. How are you envisioning toggling this on and off when the feature flag is on?

PR #1305, opened 2024-01-22
Auth 2261 enable elb access logs

pskushwaha1:
## What?

AUT-2261-AWS ELB Logging Enabled (Account)

## Why?
Enabling ElB logging to fix CPC Finding -CPC-223


pskushwaha1:
Will raise a fresh PR with rebase

PR #1297, opened 2024-01-18
AUT-2244: Send user journey to backend API when logging in during reauthentication

mattvot:
## What?

Send the user journey to the backend API when they log in during a reauthentication journey.

## Why?

We need to increment a relevant counter on the backend when the user logs in during reauthentication.

## Performance Analysis have been informed of the change

- [x] Performance Analysis have been informed of the change

## Related PRs

Please include links to PRs in other repositories relevant to this PR.
- https://github.com/govuk-one-login/authentication-frontend/pull/1290

PR #1312, opened 2024-01-23
AUT-2253: Remove `reauthenticate` property from user session once auth journey finishes

mattvot:
## What?

When the auth controller is hit, this removes the `reauthenticate` property from the users session.

## Why?

This is to ensure we don’t use it incorrectly in later journeys. Reauthenticate here indicates the RP requested that the user reauthenticate before performing an action on the RP.

## Related PRs

Please include links to PRs in other repositories relevant to this PR.

- https://github.com/govuk-one-login/authentication-frontend/pull/1290
- https://github.com/govuk-one-login/authentication-frontend/pull/1297

### DIFF
--- a/src/components/auth-code/auth-code-controller.ts
+++ b/src/components/auth-code/auth-code-controller.ts
@@ -24,7 +24,9 @@ export function authCodeGet(
       req.session.client,
       req.session.user
     );
+
     req.session.user.authCodeReturnToRP = false;
+    delete req.session.user.reauthenticate;
 
     if (!result.success) {
       throw new BadRequestError(result.data.message, result.data.code);
--- a/src/components/auth-code/tests/auth-code-controller.test.ts
+++ b/src/components/auth-code/tests/auth-code-controller.test.ts
@@ -45,6 +45,30 @@ describe("auth code controller", () => {
   });
 
   describe("authCodeGet", () => {
+    it("should remove the reauthenticate user session property", async () => {
+      const fakeAuthCodeService: AuthCodeServiceInterface = {
+        getAuthCode: sinon.fake.returns({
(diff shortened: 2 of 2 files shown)


src/components/auth-code/auth-code-controller.ts
```diff
@@ -24,8 +24,14 @@ export function authCodeGet(
       req.session.client,
       req.session.user
     );
+
     req.session.user.authCodeReturnToRP = false;
 
+    // Remove the reauthenticate property regardless of its value
```
BeckaL, line 30: So we tend not to comment in the code like this, I tend to prefer to put this in the commit message. I think we should probably have a team chat about our approach to comments at some point, so this one non-blocking, I'll let you decide whether you want to keep it or boost the commit message.

src/types.ts
```diff
@@ -75,7 +75,7 @@ export interface UserSession {
   isAccountRecoveryJourney?: boolean;
   isAccountRecoveryCodeResent?: boolean;
   accountRecoveryVerifiedMfaType?: string;
-  reauthenticate?: string;
+  reauthenticate?: string; // a truthy value represents the user in a reauthentication journey
```
BeckaL, line 78: I'd probably prefer not to comment here. Perhaps it means we should change the name of the parameter to "reauthenticateEmail". The original commit message does have some explanation of this https://github.com/govuk-one-login/authentication-frontend/commit/95c4f26d2d823e9d956ed60ab420fccc91a8c751 (arguably not the best)

BeckaL (dismissed):
Looks good to me, some non-blocking comments about comments that I'll let you decide on (but think we should discuss as devs what our approach is)

BeckaL (approved):
LGTM :-)

PR #1330, opened 2024-01-31
AUT-2306: Upgrade terraform and providers

whi-tw:
## What?

- Upgrade terraform and required_providers
- replace excluded_rule with rule_action_override
- Replace deprecated attributes with new ones
- Add helper script to lint terraform

## Why?

Terraform 1.7.1 is now being used, and we need to keep everything up to date and in sync.


scripts/terraform-lint.sh
```diff
@@ -0,0 +1,25 @@
+#!/usr/bin/env bash
```
BeckaL, line 1: Just wondering if we need this? I just run `terraform fmt --recursive` and haven't needed anything different

.git-blame-ignore-revs
```diff
@@ -0,0 +1,2 @@
+# Reformatting of deploy-authdevs.sh
```
BeckaL, line 1: Ohhh this is cool!!! Did not know you could do this, was going to ask you to squash that commit before I saw it was for changes not in this PR. ⭐

BeckaL (approved):
Looks good to me, have tested this out on sandpit. One question there about whether we need a script or not but I'm not too bothered either way if it's been useful, so approving (ping me if you change anything and want me to tick again)

.git-blame-ignore-revs
whi-tw: :D the problem is, you need a second commit to update this file, as you need to know the SHA of the commit where the formatting changes were done! I wish there was some metadata you could put in the commit message like [reformat-only] :c

PR #1335, opened 2024-02-05
Enabled drop invalid http header on frontend alb

pskushwaha1:
## What?

AUT-2270 --Ensure that Drop Invalid Header Fields feature is enabled for your Amazon Application Load Balancers (ALBs)

## Why?

If Drop Invalid Header Fields security feature is enabled, HTTP headers with header fields that are not valid are removed by the Application Load Balancer instead of being routed to the associated targets


pskushwaha1:
closing this PR will raise a new one

PR #1339, opened 2024-02-06
AUT-2032: Set up 2FA reset password flow for Password Reset Required

LazarAlexandru-Constantin:
## What?

Set up ‘Check your phone’ flow for Password Reset Required

## Why?

Users can be forced to reset their password if they previously created a weak password.  This flow needs updating too to ask for a 2FA (SMS).

### DIFF
--- a/src/components/common/state-machine/state-machine.ts
+++ b/src/components/common/state-machine/state-machine.ts
@@ -75,6 +75,8 @@ const authStateMachine = createMachine(
       isAccountRecoveryJourney: false,
       support2FABeforePasswordReset: false,
       isReauthenticationRequired: false,
+      requiresResetPasswordMFASmsCode: false,
+      requiresResetPasswordMFAAuthAppCode: false,
     },
     states: {
       [PATH_NAMES.START]: {
@@ -380,6 +382,14 @@ const authStateMachine = createMachine(
         on: {
           [USER_JOURNEY_EVENTS.CREDENTIALS_VALIDATED]: [
             {
+              target: [PATH_NAMES.RESET_PASSWORD_2FA_SMS],
+              cond: "is2FASMSPasswordChangeRequired",
+            },
+            {
+              target: [PATH_NAMES.RESET_PASSWORD_2FA_AUTH_APP],
+              cond: "is2FAAuthAppPasswordChangeRequired",
+            },
+            {
               target: [PATH_NAMES.RESET_PASSWORD_REQUIRED],
               cond: "isPasswordChangeRequired",
             },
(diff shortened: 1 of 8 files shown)


src/components/enter-password/enter-password-controller.ts
```diff
@@ -216,9 +219,10 @@ export function enterPasswordPost(
             req.session.user.isLatestTermsAndConditionsAccepted,
           requiresTwoFactorAuth: userLogin.data.mfaRequired,
           isConsentRequired: req.session.user.isConsentRequired,
-          mfaMethodType: userLogin.data.mfaMethodType,
+          mfaMethodType: req.session.user.enterEmailMfaType,
```
DTaylorGDS, line 222: I don't necessarily have an opinion on the right thing here, but why is it that this is using the session, whereas requiresTwoFactorAuth is using the API response? `mfaMethodType` does come back in the API response, so it may be safer to use that and let the backend keep track of these things. As I say, not sure I have a strong opinion, so if it's a deliberate choice, cool

src/components/enter-password/tests/enter-password-2fa-before-password-reset-required-integration.test.ts
```diff
@@ -0,0 +1,196 @@
+import request from "supertest";
```
DTaylorGDS, line 1: I'm not sure about the file name. If I understoof correctly, this file is basically, enter-password integration tests when password reset is enabled at env level. Not necessarily with 2FA (e.g. there's a test that simulates `vtr=[Cl]`).

Could we possibly rename the file to make the purpose clear and same for the reset-password one?

DTaylorGDS (commented):
Looks good in general. A couple of questions rather than suggestions from me.

src/components/enter-password/enter-password-controller.ts
LazarAlexandru-Constantin: Agreed, at this point in the journey we had the enterEmailMfa Type field populated, but there is no actual need to replace the field, as the functionality is the same. Would indeed be safer to leave the implementation as it was before.

src/components/enter-password/tests/enter-password-2fa-before-password-reset-required-integration.test.ts
LazarAlexandru-Constantin: I suggest the name enter-password-support-2fa-before-password-reset-integration.test, as this is testing the controller for when the SUPPORT_2FA_B4_PASSWORD_RESET environment variable is turned on, which implies also that the mfaRequired must be true for this test case.
This is also a naming convention that was used across the 2fa before password reset epic.

LazarAlexandru-Constantin:
After the PR for AUT-2220 to add the last 3 digits to the userExists Api was merged, the frontend needed updating, which I have done as part of this second commit in this PR. Since this was a small change, I have added this as part of this PR to finish the 2FA before password reset epic. Hope this is alright, let me know if it needs changing. 
[PR in question](https://github.com/govuk-one-login/authentication-api/pull/3900)

src/components/common/state-machine/state-machine.ts
```diff
@@ -754,6 +776,15 @@ const authStateMachine = createMachine(
         context.mfaMethodType === MFA_METHOD_TYPE.SMS &&
         context.support2FABeforePasswordReset === true,
       isPasswordChangeRequired: (context) => context.isPasswordChangeRequired,
+      is2FASMSPasswordChangeRequired: (context) =>
+        context.isPasswordChangeRequired === true &&
+        context.mfaMethodType === MFA_METHOD_TYPE.SMS &&
+        context.support2FABeforePasswordReset === true &&
+        context.requiresTwoFactorAuth === true,
```
dbes-gds, line 783: sms has context.requiresTwoFactorAuth === true but auth_app doesn't, is that correct?

LazarAlexandru-Constantin: Nice spot, thanks. Must've removed it by accident.

LazarAlexandru-Constantin: Done

dbes-gds (approved):

PR #1354, opened 2024-02-09
AUT-2093-part-1: Remove previous implementation of reauth landing screen

ayoshebby:
## What?
- This PR is the first part of creating a new screen for the reauthentication journey
- Previously, it was decided for the user to be redirected to the /enter-password when reauthentication is required
- It has now been decided that the user to be redirected to a new re-auth screen

## Why?

New re-auth screen is now available, user will be redirected to this screen instead

## Related PRs

Revert some of the Initial PR to redirect user to /enter-password when re-authentication is required
(https://github.com/govuk-one-login/authentication-frontend/pull/1262)

### DIFF
--- a/src/components/common/state-machine/state-machine.ts
+++ b/src/components/common/state-machine/state-machine.ts
@@ -129,7 +129,7 @@ const authStateMachine = createMachine(
               cond: "isConsentRequired",
             },
             {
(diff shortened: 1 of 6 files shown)


BeckaL (approved):
Looks sensible to me

PR #1358, opened 2024-02-12
AUT-2370: Wire account-intervention middleware into prove identity route

BeckaL:
Currently, the account interventions middleware is called in /auth-code. However, in cases where a user needs to re-prove their identity, this journey happens before /auth-code is hit, and we should circumvent flagged users from continuing their journey at this point.

This also corrects a small issue in the middleware where the redirects were not being returned by the function, which caused issues in tests.


src/components/prove-identity/tests/prove-identity-integration.test.ts
```diff
@@ -0,0 +1,133 @@
...
+import decache from "decache";
+import {
+  API_ENDPOINTS,
+  HTTP_STATUS_CODES,
+  PATH_NAMES,
+} from "../../../app.constants";
+import { expect } from "chai";
+import { Application } from "express";
+
+describe.only("Integration::prove identity", () => {
```
mattvot, line 14: Just spotted this .only. Best remove that before merging.

We may be best adding an eslint rule for this to the codebase, https://stackoverflow.com/a/54015366

mattvot (changes requested):

src/components/prove-identity/tests/prove-identity-integration.test.ts
BeckaL: Ahhhhhhh!!!! I thought I'd taken these out, thanks for the spot

dbes-gds (approved):

mattvot (approved):

PR #1364, opened 2024-02-14
AUT-2389: New 2FA content missing from 'Enter security code from your auth app' page

LazarAlexandru-Constantin:
## What?

Add New 2FA content missing from 'Enter security code from your auth app' page.

## Why?

New UI/UX design is required

### DIFF
--- a/src/locales/cy/translation.json
+++ b/src/locales/cy/translation.json
@@ -2197,6 +2197,18 @@
           "label": "Rhowch y cod diogelwch",
           "labelSummary": "Dyma’r rhif 6-digid a ddangosir yn eich ap dilysydd"
         }
+      },
+      "2FaBeforeResetPassword": {
+        "title": "Enter a security code from your authenticator app",
+        "header": "Enter a security code from your authenticator app",
+        "info": {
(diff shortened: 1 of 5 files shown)


src/locales/cy/translation.json
```diff
@@ -2197,6 +2197,17 @@
           "label": "Rhowch y cod diogelwch",
           "labelSummary": "Dyma’r rhif 6-digid a ddangosir yn eich ap dilysydd"
         }
+      },
+      "2FaBeforeResetPassword": {
+        "header": "Enter a security code from your authenticator app",
```
LazarAlexandru-Constantin, line 2202: This is put in English on purpose, to indicate that when we will get the welsh translation for this, we will need to replace the text with the translation.

dbes-gds: noted

dbes-gds (commented):
Let's update this as the title does not currently match the header text.

{% set pageTitleName = 'pages.enterAuthenticatorAppCode.title' | translate %}

dbes-gds (approved):

PR #1365, opened 2024-02-14
Aut 2396/call account interventions on password reset

BeckaL:
## What?

This ensures that, when a user goes through the "forgotten password" flow, they are sent to the account suspended or blocked screen instead of being able to re-set their password.

## Why?

We should not allow users to re-set their passwords who have account interventions set on them.

### DIFF
--- a/src/middleware/account-interventions-middleware.ts
+++ b/src/middleware/account-interventions-middleware.ts
@@ -7,6 +7,7 @@ import { supportAccountInterventions } from "../config";
 
 export function accountInterventionsMiddleware(
   handleSuspendedStatus: boolean,
+  handlePasswordResetStatus: boolean,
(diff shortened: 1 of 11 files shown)


src/middleware/account-interventions-middleware.ts
```diff
@@ -30,7 +32,21 @@ export function accountInterventionsMiddleware(
             USER_JOURNEY_EVENTS.PERMANENTLY_BLOCKED_INTERVENTION
           )
         );
-      } else if (accountInterventionsResponse.data.passwordResetRequired) {
+      } else if (
```
dbes-gds, line 35: Have a question about whether suspended should be before password reset.

dbes-gds (approved):

PR #1382, opened 2024-02-19
AUT-2356: Update error screens when a user has entered an incorrect SMS code more than 5 times

LazarAlexandru-Constantin:
## What?

Update the content of the error screens that appear when a user is locked out due to entering and incorrect SMS OTP too many times

## Why?

We are increasing the lockout time to 2 hours, so the lockout time needs to be accurately reflected.

This is longer than the idle session time out, so the user journey changes.

Journeys this needs to be done in:
 * Create a  One Login account- will need different content to the rest of the journeys

 * Sign in to One Login

 * Reset password

 * 2fa account recovery - when user is setting up their new 2fa method

The changes for this were demonstrated to Kim and to Craig, Kim and other colleagues in a demo.

### DIFF
--- a/src/components/account-recovery/check-your-email-security-codes/send-email-otp-middleware.ts
+++ b/src/components/account-recovery/check-your-email-security-codes/send-email-otp-middleware.ts
@@ -1,10 +1,15 @@
 import { Request, Response, NextFunction } from "express";
 import { sendNotificationService } from "../../common/send-notification/send-notification-service";
-import { JOURNEY_TYPE, NOTIFICATION_TYPE } from "../../../app.constants";
+import {
+  JOURNEY_TYPE,
+  NOTIFICATION_TYPE,
+  PATH_NAMES,
+} from "../../../app.constants";
 import { ERROR_CODES, getErrorPathByCode } from "../../common/constants";
 import { BadRequestError } from "../../../utils/error";
 import xss from "xss";
 import { ExpressRouteFunc } from "../../../types";
+import { support2hrLockout } from "../../../config";
 
 export function sendEmailOtp(
   notificationService = sendNotificationService()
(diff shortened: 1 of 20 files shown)


ayoshebby (approved):

src/components/account-recovery/check-your-email-security-codes/send-email-otp-middleware.ts
```diff
@@ -39,6 +44,18 @@ export function sendEmailOtp(
     ) {
       return res.render("security-code-error/index-wait.njk");
     }
+    if (
```
dbes-gds, line 47: This won't only show if support2hrLockout() is true so counts as a change to live.  Is that correct?

src/components/enter-password/enter-password-controller.ts
```diff
@@ -187,6 +187,8 @@ export function enterPasswordPost(
       }
     }
 
+    req.session.user.isSignInJourney = true;
```
dbes-gds, line 190: no use of the support2hrLockout() switch here, is it not required?

src/components/enter-mfa/enter-mfa-controller.ts
```diff
@@ -27,6 +27,19 @@ export function enterMfaGet(
   return async function (req: Request, res: Response) {
     const isAccountRecoveryEnabledForEnvironment = supportAccountRecovery();
 
+    if (
```
dbes-gds, line 30: This won't only show if support2hrLockout() is true so counts as a change to live. Is that correct?

src/components/enter-password/enter-password-controller.ts
LazarAlexandru-Constantin: the session variables for the journeys are only used now in the 2hr screens, I didn't put them them in a flag as they wouldn't impact anything else (as they are new and not used anywhere else) and they might me useful in other tickets

src/components/enter-mfa/enter-mfa-controller.ts
LazarAlexandru-Constantin: My intuition is that this code should be here, but I'll wrap it in a feature switch just to be sure

src/components/account-recovery/check-your-email-security-codes/send-email-otp-middleware.ts
LazarAlexandru-Constantin: My intuition is that this code should be here, but I'll wrap it in a feature switch

LazarAlexandru-Constantin:
Hi Gwyn, yes thanks for the remark. I have updated the PR description. We had a demo yesterday for these changes and did a separate demo with Kim also to double check.

dbes-gds (approved):

PR #1386, opened 2024-02-19
AUT-2417: add lockout logic for reset passwd 2fa auth app controller

LazarAlexandru-Constantin:
## What?

add lockout logic for reset-password-2fa-auth-app-controller.ts

## Why?

Enter a security code from your auth app - 6 incorrect codes - user can immediately navigate back to auth app screen without waiting for lockout period to expire


mark-cab:
don't we use the state machine optional paths to manage these?

PR #1392, opened 2024-02-20
BAU: Add `dev:render-user-journey-documentation` command

mattvot:
## What?

Create a dev only command `yarn dev:render-user-journey-documentation` that generates temporary local documentation derived from the user journey state machine.

A HTML table is created that has had the headers "Current Path", "Input Event", "Next Path & "Optional Paths" and can be used to understand how a user can navigate through their journey.

Definitions of the conditions to get to the next paths are added inline with the `title` attribute and the fall through nature of next paths are visualised.

## Why?

Understanding how the user journey is implemented can take a while when looking at `src/components/common/state-machine/state-machine.ts`. This documentation should reduce the time to understand it.

https://github.com/govuk-one-login/authentication-frontend/assets/723372/c7822553-54a4-4a06-940e-deb5c30e55a2

### DIFF
--- a/README.md
+++ b/README.md
@@ -166,6 +166,14 @@ If you're having problems running locally, try these steps first:
 
 Remember to run these commands in the docker container itself.
 
+### Documentation
+
+> Generate and view documentation of the user journey state machine 
+
+```shell script
+yarn dev:render-user-journey-documentation
+```
+
 ### Development
 
 > To run the app in development mode with nodemon watching the files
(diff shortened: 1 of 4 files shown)


DTaylorGDS (dismissed):

DTaylorGDS (approved):

BeckaL (approved):
Thanks for this! LGTM

PR #1413, opened 2024-02-27
AUT-2441: Fixed journey to end on 'password reset required' when user changes how they get security codes with resetPasswordRequired and temporarilySuspended AIS statuses applied

alhcomer:
## What?

Fixed journey so that when a user has the reset password and temporarily suspended statuses applied to their account and they change how they get security codes they are now taken to the 'reset password required' screen instead of the 'temporarily suspended screen'

## Why?

Users were incorrectly being taken to the temporarily suspended page.


src/components/common/state-machine/state-machine.ts
```diff
@@ -725,6 +725,9 @@ const authStateMachine = createMachine(
           [USER_JOURNEY_EVENTS.EMAIL_SECURITY_CODES_CODE_VERIFIED]: [
             PATH_NAMES.GET_SECURITY_CODES,
           ],
+          [USER_JOURNEY_EVENTS.PASSWORD_RESET_INTERVENTION]: [
```
BeckaL, line 728: Could you add a test for this? I think you should be able to modify one of the existing ones in the reset password integration tests. Also, would we not need to modify the middleware config too? (the true / true thing we were talking about yesterday?) If easier to chat through on a call let me know

src/components/common/verify-code/verify-code-controller.ts
```diff
@@ -104,6 +104,9 @@ export function verifyCodePost(
         if (accountInterventionsResponse.data.temporarilySuspended) {
           nextEvent = USER_JOURNEY_EVENTS.TEMPORARILY_BLOCKED_INTERVENTION;
         }
+        if (accountInterventionsResponse.data.passwordResetRequired) {
```
BeckaL, line 107: I don't follow why we need this? Is this covering off a different bug to the password reset one we were talking about yesterday? Could we have a test for this?

BeckaL (commented):
Some comments / questions - let me know if easier to chat through on a call, I might be misremembering some things

src/components/common/state-machine/state-machine.ts
BeckaL: Ah sorry I was thinking of the other ticket we were on, so ignore the middleware comment. I still would want an overall test for this

src/components/common/verify-code/verify-code-controller.ts
BeckaL: Same as above, sorry was entirely misinterpreting this as being the work we talked about yesterday. Test would be good

BeckaL (approved):
Code change look good to me, would just test this on sandpit out of caution if you haven't already but approving

PR #1419, opened 2024-02-29
AUT-2446: Fixed AIS Password Reset Required Journey to Allow for Password Reset Instead of Showing Temporarily Suspended Screen

alhcomer:
## What?

Fixed AIS password reset required journey to allow for password reset journey rather than redirecting to temporarily suspended screen.

## Why?

After password reset required screen, entering OTP and MFA a user with passwordResetRequired and temporarilySuspended statuses was incorrectly being shown the temporarily suspended screen instead of being taken through the password reset journey.

### DIFF
--- a/src/components/common/verify-code/verify-code-controller.ts
+++ b/src/components/common/verify-code/verify-code-controller.ts
@@ -100,12 +100,10 @@ export function verifyCodePost(
           );
         if (accountInterventionsResponse.data.blocked) {
           nextEvent = USER_JOURNEY_EVENTS.PERMANENTLY_BLOCKED_INTERVENTION;
-        }
-        if (accountInterventionsResponse.data.temporarilySuspended) {
-          nextEvent = USER_JOURNEY_EVENTS.TEMPORARILY_BLOCKED_INTERVENTION;
-        }
-        if (accountInterventionsResponse.data.passwordResetRequired) {
+        } else if (accountInterventionsResponse.data.passwordResetRequired) {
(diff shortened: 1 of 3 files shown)


src/components/common/verify-code/verify-code-controller.ts
```diff
@@ -100,12 +100,10 @@ export function verifyCodePost(
...
         if (accountInterventionsResponse.data.blocked) {
           nextEvent = USER_JOURNEY_EVENTS.PERMANENTLY_BLOCKED_INTERVENTION;
-        }
-        if (accountInterventionsResponse.data.temporarilySuspended) {
-          nextEvent = USER_JOURNEY_EVENTS.TEMPORARILY_BLOCKED_INTERVENTION;
-        }
-        if (accountInterventionsResponse.data.passwordResetRequired) {
+        } else if (accountInterventionsResponse.data.passwordResetRequired) {
           nextEvent = USER_JOURNEY_EVENTS.PASSWORD_RESET_INTERVENTION;
+        } else if (accountInterventionsResponse.data.temporarilySuspended) {
```
mark-cab, line 105: Can you confirm intervention takes priority over suspended accounts?

alhcomer: Password reset intervention takes priority over temporary suspension as all accounts with password reset required are temporarily suspended, though not all temporarily suspended accounts have password reset required. Meaning a user must have password reset intervention removed first before then re-running the check to see if temporary suspensions are applied for other reasons

BeckaL (approved):

PR #1441, opened 2024-03-07
BAU: Make docker-compose listening ports easily editable

whi-tw:
### make dev docker exposed ports modifiable

Rather than the ports being hardcoded in `docker-compose.yml`, instead set them via environment variables.

This means people can change the ports locally in a non-committed file in the case of a clash (an existing service listening on the same port)

### YAML formatting in `docker-compose.yml`

Rather than using the `list` format for environment variables, instead define them as key:value pairs.

This makes the key and the value visually distinct with syntax highlighting enabled.

### Add checks for required environment variables in `docker compose`

Rather than just showing a warning for missing variables, instead error when required ones are not defined.

This means that the frontend can only be started when all the environment variables are set, reducing confusion
in the case that a required one is not set, and it's not obvious that a missing envar is the cause.

### Add tests for required node environment variables

Some values are not provided in the samples, and need to be added. Weird stuff happens when they're unset.

As we know that missing these will cause problems, we can check them in advance.

Additionally, this gives us a good place to test if the `$ORCH_TO_AUTH_SIGNING_KEY` value is correctly formatted - it cannot be set with `\n`, but has to have actual newline characters to be processed correctly.

## How to review

> NB: ensure that you keep a copy of your `.env` file, so it's not lost forever!

1. Ensure the readme changes are sensible / readable.
1. remove your `.env` file
1. run `startup.sh`: note that relevant errors are thrown about missing envars.
1. restore the `.env` file
1. run `startup.sh`: note that the errors go away and the startup completes correctly
1. run `docker compose up`: note that the startup fails due to missing AWS variables. note that the error message explains how this should be set
1. change one of the `DOCKER_.*_PORT` variables in `.env`
1. run `startup.sh`: note that the service whos port you changed is now listening on the new port
1. ensure that the service is accessible on that new port


README.md
```diff
@@ -60,7 +50,7 @@ The startup script will do this for you so just run this command:
 
 When starting for the first time, or after a clean, the frontend will take a few minutes to start as node needs to install all the dependencies.
 
-To find out if the application has started, open a console window on the frontend docker container and view the logs. If the server has started successfully you will see this message `Server listening on port 3000`.  If this does not appear try forcing node to restart by updating one of the `.njk` files.
+To find out if the application has started, open a console window on the frontend docker container and view the logs. If the server has started successfully you will see this message `Server listening on port 3000`. If this does not appear try forcing node to restart by updating one of the `.njk` files.
```
gtvj, line 53: Is it worth explaining here why a change will cause Node to restart. Something along the lines of 

> ...because Node.js is watching files, any changes to files will cause Node.js to restart. If you don't see the `Server listening on port 3000` message, you can force a restart by running `touch src/server.ts` as a Terminal command or manually editing a file in the `src` directory

README.md
```diff
@@ -21,20 +21,10 @@ Before you can run the frontend app against the backend you will need to configu
 
 ### Set the Environment variables
 
-Create a copy of the .env.sample file and rename it .env and fill in below values:
+Create a copy of one of the .env.XXX files (it's best to start with the one named the same as your target environment),
```
gtvj, line 24: Hi @whi-tw - sorry if this is a silly question, but how would I know what my target environment is? I'm assuming it might typically be build (although I might be wrong about that). I'm just wondering if it might be helpful to let people know where they'd find out what their environment is.

gtvj (commented):
Hi @whi-tw - I'm not too familiar with the Docker stuff included here but have left a couple of comments.

gtvj:
Hi @whi-tw, 

I've started running through the "How to review" section and hit an error when trying to run Step 5

> No local running node processes stopped...
> ERROR: $AWS_PROFILE is not set. This should be exported within the script that sources this file.

README.md
whi-tw: 100% agree, but that should be a separate PR

README.md
whi-tw: I had trouble with the wording for this.
I think it's supposed to mean:
> if you're running this against OIDC in `build` use `.env.build` as the template

whi-tw:
> Hi @whi-tw,
> 
> I've started running through the "How to review" section and hit an error when trying to run Step 5
> 
> > No local running node processes stopped...
> > ERROR: $AWS_PROFILE is not set. This should be exported within the script that sources this file.

Your `.env` file doesn't have `AWS_PROFILE` set!

src/config.ts
```diff
@@ -112,7 +112,9 @@ export function getKmsKeyId(): string {
 }
 
 export function getOrchToAuthSigningPublicKey(): string {
-  return process.env.ORCH_TO_AUTH_SIGNING_KEY;
+  const publicKey = process.env.ORCH_TO_AUTH_SIGNING_KEY;
```
dbes-gds, line 115: this is a production change, need to keep an eye in the pipeline.

startup.sh
```diff
@@ -1,71 +1,78 @@
...
-  docker compose -f "docker-compose.yml" up -d --wait --no-deps redis di-auth-stub-default di-auth-stub-no-mfa
-  export REDIS_PORT=6389
+  docker compose -f docker-compose.yml up --build -d --wait
+  echo "No-MFA stub listening on http://localhost:${DOCKER_STUB_NO_MFA_PORT}"
+  echo "Default stub listening on http://localhost:${DOCKER_STUB_DEFAULT_PORT}"
+  echo "Redis listening on redis://localhost:${DOCKER_REDIS_PORT:-6379}"
+  export REDIS_PORT=${DOCKER_REDIS_PORT:-6379}
   export REDIS_HOST=localhost
-  yarn install && yarn copy-assets && yarn dev
+  if [ $DEPS_ONLY == "0" ]; then
```
dbes-gds, line 65: server is not starting for me with just the -l option as the yarn commands are in here and aren't running

whi-tw: Odd. I can't reproduce this at all - `./startup.sh -l` results in the `yarn` tasks running, `./startup.sh -lx` results in just the docker bits running (and then the container logs being tailed).

Can you please confirm that you're definitely passing the correct options?

Slight rework incoming to tighten up the `if` statements.

src/config.ts
whi-tw: 100% agree. I added unit tests to ensure that the behaviour is the same for both with `\n` and with actual newline characters (`0x0a`).

The alternative here is to define the certs as multiline variables in the `.env.x` files:

```bash
ORCH_TO_AUTH_SIGNING_KEY="-----BEGIN PUBLIC KEY-----
MFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAENB3csRUIdoaTHNn079Jl7JpiXzxF
0p2ZIddCErxtIhGMTTqtbQZJCPesSKUVE/DQbpIko3mLoisuFgmQfFouCw==
-----END PUBLIC KEY-----"
```

This way the environment variable is explicitly configured the same way as it is in the running environments. However, it’s a multiline bash variable and that has the potential to trip people up, so defining with `\n` seems overall safer.

Thoughts?

dbes-gds (approved):

PR #1443, opened 2024-03-08
AUT-2543: Created isForcedPasswordReset parameter in updatePassword method

alhcomer:
## What
Created isForcedPasswordReset parameter in updatePassword method to accept req.session.user.withinForcedPasswordResetJourney as an argument in the reset password controller. This is so that in the backend we can distinguish between password reset journey types for auditing.

Another PR is to follow for the backend.

## How to review

Code should be reviewed alongside backend PR (https://github.com/govuk-one-login/authentication-api/pull/4087) ensuring that the argument for isForcedPasswordReset is accessible in the ResetPasswordHandler

## Related PRs
https://github.com/govuk-one-login/authentication-api/pull/4087

### DIFF
--- a/src/components/reset-password/reset-password-controller.ts
+++ b/src/components/reset-password/reset-password-controller.ts
@@ -39,7 +39,7 @@ export function resetPasswordPost(
   mfaCodeService: MfaServiceInterface = mfaService()
 ): ExpressRouteFunc {
   return async function (req: Request, res: Response) {
-    const { email } = req.session.user;
+    const { email, withinForcedPasswordResetJourney } = req.session.user;
     const { sessionId, clientSessionId, persistentSessionId } = res.locals;
     const newPassword = req.body.password;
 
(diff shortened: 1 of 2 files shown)


src/components/reset-password/reset-password-controller.ts
```diff
@@ -48,7 +48,8 @@ export function resetPasswordPost(
       req.ip,
       sessionId,
       clientSessionId,
-      persistentSessionId
+      persistentSessionId,
+      req.session.user.withinForcedPasswordResetJourney
     );
```
ayoshebby, line 53: Minor suggestion, I think for readability you can extract this into its own var 
```
const {user, withinForcedPasswordResetJourney} = req.session.user
```
replacing req.session.user.withinForcedPasswordResetJourney >> with withinForcedPasswordResetJourney

ayoshebby (approved):

PR #1445, opened 2024-03-08
AUT-2552: ignore reset password intervention within middleware when password has been more recently reset than intervention applied

BeckaL:
This fixes a problem whereby when someone resets their password, we redirect them to /auth-code and immediately call the interventions middleware again, at which point we will prompt them to reset their password again, since this does not give enough time for the account interventions service to be updated.

Instead here, we:
* set the timestamp of the time when a user reset their password on the session
* read from this value in the middleware when we have the `resetPassword` intervention set. If a) there is a resetPasswordTime in the session, and b) it is later than the `appliedAt` time on the intervention, we ignore the intervention and continue the journey.

This also contains some refactoring commits to make tests easier to read.

Note that this alone will not be enough to ensure that a user who has reset their password can continue on with their journey outside of auth: we will also need to pass this information back to orchestration who will need to implement similar logic on their side. This will be handled in a [separate ticket](https://govukverify.atlassian.net/browse/AUT-2553).

## How to review

1. Code review
2. Set yourself up in build with a `resetPasswordRequired` and `suspended` intervention. Run this locally against build's backend, and perform a sign in journey. Once you've gone through password reset, you should not be prompted to password reset again, and should instead go through to the next stage of the journey (it is likely that you'll see a failure after the /auth-code stage because of how the local setup works). Compare this with the behaviour on `main`.
3. Note that when you reset your password in build, the timestamp at which you reset your password should always be later than the timestamp the intervention is applied at, since we [hardcode this in the stub](https://github.com/govuk-one-login/authentication-api/blob/main/interventions-api-stub/src/main/java/uk/gov/di/authentication/interventions/api/stub/entity/InterventionsApiStubResponse.java#L61) to a timestamp that equates to `Mon Oct 09 2023 16:30:05`. If we wanted at some point we could add these timestamps to the stub database to simulate alternative behaviour.

### DIFF
--- a/src/components/account-recovery/check-your-email-security-codes/tests/check-your-email-security-codes-controller.test.ts
+++ b/src/components/account-recovery/check-your-email-security-codes/tests/check-your-email-security-codes-controller.test.ts
@@ -14,7 +14,10 @@ import {
   RequestOutput,
   ResponseOutput,
 } from "mock-req-res";
-import { accountInterventionsFakeHelper } from "../../../../../test/helpers/account-interventions-helpers";
+import {
+  accountInterventionsFakeHelper,
+  noInterventions,
+} from "../../../../../test/helpers/account-interventions-helpers";
 import { fakeVerifyCodeServiceHelper } from "../../../../../test/helpers/verify-code-helpers";
 import { ERROR_CODES } from "../../../common/constants";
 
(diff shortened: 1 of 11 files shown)


alhcomer (approved):
Ran through in build as outlined in the PR and the changes successfully resolve looping password reset issue. All changes look good :)

PR #1463, opened 2024-03-19
AUT-2612: Reorder validation messages on create password screen

gtvj:
## What

UCD identified a bug today relating to the order of validation messages when a user leaves both password fields blank during an account creation journey. The problem is that "Re-type your password" appears before "Enter your password" (whereas the fields are in the opposite order). This PR changes the order of these messages. 

I have also checked that the page retains existing validation behaviour, including:
- [x] failures relating to using a common password, like "password1"
- [x] leaving a single field blank
- [x] using a password that doesn't include numbers
- [x] allows submission when a good password is used to populate both fields

## How to review

1. Start frontend running locally against this branch
2. Begin an account creation journey
3. On the "Create your password", leave both fields blank and submit the form
4. Confirm that the error message order matches what's in the screenshots below
5. Try different validation error permutations to confirm all is working as expected
6. Code review

### Screenshots

<details>
<summary>English</summary>

Error-Create-your-password-GOV-UK-One-Login

</details>

<details>
<summary>Welsh</summary>

Gwall-Creu-eich-cyfrinair-GOV-UK-One-Login

</details>

### DIFF
--- a/src/components/create-password/create-password-validation.ts
+++ b/src/components/create-password/create-password-validation.ts
@@ -5,24 +5,6 @@ import { ValidationChainFunc } from "../../types";
 
 export function validateCreatePasswordRequest(): ValidationChainFunc {
   return [
-    body("confirm-password")
-      .notEmpty()
-      .withMessage((value, { req }) => {
-        return req.t(
-          "pages.createPassword.confirmPassword.validationError.required",
-          { value }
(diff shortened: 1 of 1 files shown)


BeckaL (approved):
Sorry this has been hanging round for a bit!

PR #1469, opened 2024-03-20
Aut 2623/allow actions for reprove identity

BeckaL:
Have made changes that allow for a user to either reset their password or change their security codes if they have an AIS suspension status placed against their account while also requiring that they reprove identity. Now they are allowed to to reset their password or change their MFA method before being taken on to the reprove identity process where they will continue their journey.

## How to review

1. Code Review
2. Run the front end against build with reprove identity and suspension statuses against your account
3. Go through the normal (non-forced) password reset journey successfully
4. Go through the change MFA journey successfully

### DIFF
--- a/src/components/common/verify-code/verify-code-controller.ts
+++ b/src/components/common/verify-code/verify-code-controller.ts
@@ -17,10 +17,8 @@ import {
   support2FABeforePasswordReset,
   supportAccountInterventions,
 } from "../../../config";
-import {
-  AccountInterventionsInterface,
-  AccountInterventionStatus,
-} from "../../account-intervention/types";
+import { AccountInterventionsInterface } from "../../account-intervention/types";
+import { isSuspendedWithoutUserActions } from "../../../utils/interventions";
 
 interface Config {
   notificationType: NOTIFICATION_TYPE;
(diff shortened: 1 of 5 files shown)


src/components/common/verify-code/verify-code-controller.ts
```diff
@@ -114,7 +117,10 @@ export function verifyCodePost(
           if (options.journeyType !== JOURNEY_TYPE.PASSWORD_RESET_MFA) {
             nextEvent = USER_JOURNEY_EVENTS.PASSWORD_RESET_INTERVENTION;
           }
-        } else if (accountInterventionsResponse.data.temporarilySuspended) {
+        } else if (
+          accountInterventionsResponse.data.temporarilySuspended &&
```
BeckaL, line 121: Just a note that you don't need this line anymore since it's checked in the below function :-) 

I also wonder about moving the "isSuspendedWithoutUserActions" to somewhere shared so we don't have two duplicated functions. Happy to have a chat about where we think that shared function should go if you don't have any immediate instinct

BeckaL (commented):
Great stuff, I'm happy with this! I'll ping this over to someone who can officially sign this off

ayoshebby (approved):

PR #1473, opened 2024-03-21
ATO-450: Create a new endpoint for rp auth codes

ethanmills:
## What

Replace the dual functionality of the `/auth-code` endpoint with a new `/rp-auth-code`. 

Currently, `/auth-code` either generates a code to return the user to Orchestration, or to return to the RP if they have completed identity. This seems to be causing some hard to diagnose issues with users occasionally being directed to the wrong place. Splitting the endpoints allows us to make better use of the state machine to make sure the user ends up in the right place, even if they use the browser back button.

Changes to the user interface or content should be demonstrated to Content Design and Interaction Design before being merged. This is to ensure they are aware of changes and behaviours (for example, how error screens appear and whether invalid entries can be amended), and can make any necessary changes to Figma.

- [ ] Changes to the user interface have been demonstrated

### DIFF
--- a/src/components/rp-auth-code/rp-auth-code-controller.ts
+++ b/src/components/rp-auth-code/rp-auth-code-controller.ts
@@ -1,15 +1,15 @@
 import { Request, Response } from "express";
 import { ExpressRouteFunc } from "../../types";
-import { authCodeService } from "./auth-code-service";
-import { AuthCodeServiceInterface } from "./types";
+import { rpAuthCodeService } from "./rp-auth-code-service";
+import { RpAuthCodeServiceInterface } from "./types";
 import { BadRequestError } from "../../utils/error";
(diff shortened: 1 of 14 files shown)


src/components/rp-auth-code/rp-auth-code-controller.ts
```diff
@@ -0,0 +1,52 @@
...
+    const gaId = req.session.client.crossDomainGaTrackingId;
+
+    const result = await service.getAuthCode(
+      sessionId,
+      clientSessionId,
+      req.ip,
+      persistentSessionId
+    );
+
+    req.session.user.authCodeReturnToRP = false;
```
hannah-leong, line 26: Should this be here?

alhcomer:
Closing the PR for the time being, left the branch as is for when @ethanmills wants to pick up again

PR #1477, opened 2024-03-22
AUT-2617: add the correct journey type to resend-mfa-code-controller.ts

LazarAlexandru-Constantin:
To implement the correct locking for code requests, we need to send the correct journey type to the backend API. Because the resend-mfa-code-controller.ts is used in the creation, account-part-created and account recovery journeys, we need to make sure that we send the current journey type, not JOURNEY_TYPE.REGISTRATION. Backend implements locking on the 6th requests. The enter-phone-controller sent the correct journey type, while the resend-mfa-code-controller.ts sent the wrong one, reason why we were blocked on the 6th, not 5th request. Obs: part-created and creation are using JOURNEY_TYPE.REGISTRATION, while account recovery has its own journey type.

## What

Update resend-mfa-code-controller.ts to send correct journey type

## How to review

1. Deploy to sandpit with ./deploy-sandpit.sh -l or test locally by staring the service with './startup.sh -l'
2. Sign in to existing account, or create a new one if you do not have one
4. Sign in
5. On the enter-code (auth app or sms), click 'change how you get security codes"
7. Enter email OTP
8. Select SMS option and enter a correct phone number
9. On the verify phone number page, request 5 codes and verify that on the 5th request you get locked out.

- [] Changes to the user interface have been demonstrated

### DIFF
--- a/src/components/account-creation/resend-mfa-code/resend-mfa-code-controller.ts
+++ b/src/components/account-creation/resend-mfa-code/resend-mfa-code-controller.ts
@@ -54,7 +54,11 @@ export const resendMfaCodePost = (
 ): ExpressRouteFunc => {
   return async function (req: Request, res: Response) {
     const { sessionId, clientSessionId, persistentSessionId } = res.locals;
-    const { email, phoneNumber } = req.session.user;
+    const { email, isAccountRecoveryJourney, phoneNumber } = req.session.user;
+
+    const journeyType = isAccountRecoveryJourney
+      ? JOURNEY_TYPE.ACCOUNT_RECOVERY
+      : JOURNEY_TYPE.REGISTRATION;
 
     const sendNotificationResponse = await service.sendNotification(
       sessionId,
@@ -64,7 +68,7 @@ export const resendMfaCodePost = (
       req.ip,
       persistentSessionId,
       xss(req.cookies.lng as string),
-      JOURNEY_TYPE.REGISTRATION,
+      journeyType,
       phoneNumber
     );
 
--- a/src/components/enter-phone-number/enter-phone-number-controller.ts
+++ b/src/components/enter-phone-number/enter-phone-number-controller.ts
@@ -84,7 +84,9 @@ export function enterPhoneNumberPost(
             req.query.actionType as SecurityCodeErrorType
           ),
           support2hrLockout: support2hrLockout(),
-          isAccountCreationJourney: true,
+          isAccountCreationJourney:
+            req.session.user.isAccountCreationJourney ||
+            req.session.user.isAccountPartCreated,
         });
       }
       const path = getErrorPathByCode(sendNotificationResponse.data.code);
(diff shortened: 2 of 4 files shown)


src/components/account-creation/resend-mfa-code/resend-mfa-code-controller.ts
```diff
@@ -54,7 +54,23 @@ export const resendMfaCodePost = (
...
+      email,
+      isAccountRecoveryJourney,
+      isAccountRecoveryPermitted,
+      phoneNumber,
+    } = req.session.user;
+
+    const isAccountRecoveryEnabledForEnvironment = supportAccountRecovery();
+
+    const accountRecovery =
+      isAccountRecoveryJourney &&
```
BeckaL, line 67: A question here because I'm not super familiar with this bit of the domain - is `isAccountRecoveryJourney` enough here? Presumably we'd only set it if account recovery was permitted and enabled for the environment?

src/components/enter-phone-number/enter-phone-number-controller.ts
```diff
@@ -84,7 +84,9 @@ export function enterPhoneNumberPost(
             req.query.actionType as SecurityCodeErrorType
           ),
           support2hrLockout: support2hrLockout(),
-          isAccountCreationJourney: true,
+          isAccountCreationJourney:
+            req.session.user.isAccountCreationJourney ||
+            req.session.user.isAccountPartCreated,
```
BeckaL, line 89: This looks like a separate change / fix to me - is that right? If so, could we either put it in its own commit with tests and explanation of what it does, or add something to the description of this commit that explains why this was changed? In the commit message for this commit you say "The enter-phone-controller sent the correct journey type", but this change implies otherwise :-)

LazarAlexandru-Constantin: Will do, agreed

src/components/account-creation/resend-mfa-code/resend-mfa-code-controller.ts
LazarAlexandru-Constantin: All of the code from this is basically copied from the enterPhoneNumberPost() controller. There this is the way that we determine what journey type we are in and pass it to the api (both enter-phone and resend-mfa-code) are used in both journeys: account-create and  account recovery
So since I would want the data that is sent to the API to be identical with the one from the  enterPhoneNumberPost controller, I did it the same way here. 

The reason for why I want to make sure the data is the same is because backend blocks users on the 6th resend-code request, and the first request of those 6 is always made by the enter phone, leaving 5 request for the resend-mfa-codes. If the data differs from the call in the enter-phone to the one in the resend-code, the resend-code will have 6 resends available because the backend will consider the requests as separate

BeckaL: I guess my thoughts are that we only set `isAccountRecovery` to true here: https://github.com/govuk-one-login/authentication-frontend/blob/main/src/components/account-recovery/check-your-email-security-codes/check-account-recovery-middleware.ts#L11-L12 where we already check whether account recovery is permitted, so I'd argue that line here is redundant. 

Checking whether it's enabled for the environment is technically still relevant so fine to leave in. It is switched on in all environments however, so looks like a flag that we can take out, but yeah while that feature flag still exists makes sense to leave it in here.

Not saying the code is wrong in either place, just that some of this seems redundant as checks to me :-)

LazarAlexandru-Constantin: Indeed, I agree with that

BeckaL (approved):

PR #1478, opened 2024-03-25
AUT-2608: send the correct flags to the .njk error pages

LazarAlexandru-Constantin:
In the reset-password-check-email-controller.ts, depending on the error codes, we redirect to 3 different error page indexes, in the block starting on line #64. Some of the indexes use the show2HrScreen, others the support2hrLockout. Thus, we need to send both of them for all the indexes to work.

## What

Implement redirect to error page when user is locked out due to too many incorrect codes entered.

## How to review

1. Code Review
2. Ensure environment variables 'SUPPORT_2HR_LOCKOUT' and 'SUPPORT_2FA_B4_PASSWORD_RESET' are set to 1
3. Deploy to sandpit with `./deploy-sandpit.sh -l`
4. for the first issue proceed through PW reset journey and enter too many wrong OTP emails and check that the screen you're redirected to after the 5th incorrect attempt is the first screenshot below
5. Restart the journey by going to localhost:2000, or going through the stub again, go to the reset-password journey and verify that you're redirected to the 2nd screenshot below. 

Expected 2hr screens to be redirected to during the PW reset journey, when entering too many wrong codes:

image

Expected 2hr screens to be redirected to during the sign-in, PW reset and account recovery journeys, when user is locked and tries to complete the journey: 
image

Expected 2hr screens to be redirected to during the creation and account part created journeys, when requesting too many codes:
image

The changes were demonstrated in a demo to Kim Clayden


BeckaL (approved):
Approving so we can get this fixed, let's create a ticket to use the same flag everywhere :-)

PR #1486, opened 2024-03-26
BAU: Add troubleshooting steps to README

gtvj:
## What

Adds a troubleshooting section that's split into general steps and how to get past specific errors.

## How to review

1. Review the new instructions
2. Review the way the instructions are presented to see if you think this works (for example, I've used a disclosure element for the steps relating to specific errors)

### DIFF
--- a/README.md
+++ b/README.md
@@ -139,6 +139,29 @@ You may need to install pre-commit for the script to work.
 brew install pre-commit
 ```
 
+## Troubleshooting the local run
+
+### General steps to try first
+
+If you're having problems running locally, try these steps first:
+
+* Connect to the VPN
+* Run `./shutdown.sh`
+* Delete your Docker **Images** (you can do this via Docker Desktop or with `docker system prune --all`)
+* Run `./startup.sh -lc` to do a cleanup before a local run
+* Because things sometimes don't work first time round, a `touch src/server.ts` _while the server is running_ might help
+
+### Getting past specific errors
+
+<details>
+
+<summary>`Error: secret option required for sessions`</summary>
+
+1. stop the server with `./shutdown.sh`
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -139,6 +139,39 @@ You may need to install pre-commit for the script to work.
...
+1. stop the server with `./shutdown.sh`
+2. run `./startup.sh -l` (there's no need for the -c flag)
+
+</details>
+
+<details>
+
+<summary>`Error: Error decrypting JWE`</summary>
+
+1. Go to `scripts/export_aws_creds.sh` and add `aws s3 ls --profile [profile-name]` as the last line to this file (replacing `[profile-name]`) with the profile from your `./startup.sh` file)
```
BeckaL, line 169: I'm just wondering if there's a less hacky solution to hardcoding the profile / adding this line in ad-hoc to something that's not committed

BeckaL (approved):
Generally approving since I think good to get this stuff in the readme. Just got a question about whether there's a nicer solution to this, but if it'll take a bit longer to sort out happy to just get this merged and iterate on this :-)

README.md
gtvj: I'm sure there probably is. To be honest, I couldn't see why adding an `ls` would change anything - unless it's initiating a connection of some sort that's relied upon later. I think this is something @whi-tw might have mentioned too.  I'd benefit from some advice from someone that has a better understanding of the infrastructure and AWS configuration.

@whi-tw - what's your view on this?

gtvj: As discussed on Slack, we're going to remove this. Just pushing an update to the commit now.

whi-tw (approved):
LGTM!

PR #1492, opened 2024-03-26
Aut 2598/add mfa method when resetting password

BeckaL:
## What

Currently, if a user:
* locks themselves out of their account due to too many passwords and then;
* attempts to access the service again in a new session and;
* chooses to reset their password when they hit the blocked screen

They will not be prompted to enter their 2fa during the password reset journey. This is because we determine whether someone needs to enter 2fa depending on whether the enterEmailMfaType is set in the session. This is normally set by the response from the "CheckUserExists" api endpoint, but when someone is locked out, this response returns an error and we do not get the mfa detail.

This commit forms part of a fix for this. We've updated the resetPasswordRequest response to contain the user's mfa details. Therefore, in this commit we:

* Add a new response interface for this request to reflect the data coming back in the body
* When we don't already have the mfa information set in the session, we set it here from the data contained in the body.

This will mean that, when a user enters the code from their email, the state machine will return the enter mfa code pages next, rather than going straight to the page where a user resets their password.

## How to review

1. Code Review
2. Run this code locally against authdev1, and run through the steps in [the ticket](https://govukverify.atlassian.net/jira/software/c/projects/AUT/boards/477?assignee=712020%3Ad3baca0d-e361-43cf-9972-bc82fc554739&assignee=620a44d9a715c600691222f8&selectedIssue=AUT-2598). Verify that this issue is fixed.

## Related PRs

The following changes to be merged first, in this order:
* **MERGED** [PR in frontend](https://github.com/govuk-one-login/authentication-frontend/pull/1491) which changes possible status codes for the reset password service 
* **MERGED** [PR in authentication-api](https://github.com/govuk-one-login/authentication-api/pull/4162) which updates the api to return the mfa details in the reset password request response

### DIFF
--- a/src/components/reset-password-check-email/reset-password-check-email-controller.ts
+++ b/src/components/reset-password-check-email/reset-password-check-email-controller.ts
@@ -53,6 +53,11 @@ export function resetPasswordCheckEmailGet(
       );
     }
 
+    if (result.success && req.session.user.enterEmailMfaType === undefined) {
+      req.session.user.enterEmailMfaType = result.data.mfaMethodType;
+      req.session.user.redactedPhoneNumber = result.data.phoneNumberLastThree;
+    }
+
     if (!requestCode || result.success) {
       const support2FABeforePasswordResetFlag = support2FABeforePasswordReset();
       return res.render(TEMPLATE_NAME, {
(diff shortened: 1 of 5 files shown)


LazarAlexandru-Constantin:
Code review done and tested locally, everything works as expected.

LazarAlexandru-Constantin (approved):

dbes-gds (approved):

PR #1497, opened 2024-03-27
Bau/remove cy feature flag

BeckaL:
## What

Removes a feature flag that has been enabled in all environments for a while, and associated logic.

Note that for production, this variable is currently set through a secret in aws, so there is no associated removal here. However, you can see in aws for the ecs instance that the value is set to 1 in production.

## How to review

1. Code Review


src/components/prove-identity-welcome/index-existing-session.njk
```diff
@@ -17,24 +17,18 @@
...
-                <span lang="{{ 'pages.proveIdentityWelcome.section1.insetAlternativeLanguage.linkText.destinationLanguageCode' | translate }}">{{ 'pages.proveIdentityWelcome.section1.insetAlternativeLanguage.linkText.inDestinationLanguage' | translate }}</span></a>.
-            {{ 'pages.proveIdentityWelcome.section1.insetAlternativeLanguage.end' | translate }}
-        {% endset %}
-
-        {{ govukInsetText({
-            html: insetTextInnerHtml
-        }) }}
-    {% else %}
-        {{ govukInsetText({
-            text: 'pages.proveIdentityWelcome.section1.insetTextEnglishOnly' | translate
```
gtvj, line 35: Just thinking we should probably remove this value from the translation files too (I think it's on line 2221 of both `cy/translation.json` and `en/translation.json`)

gtvj (changes requested):
Hi @BeckaL 👋🏽 , this looks good - just a small suggestion about removing an associated entry from the locale files.

gtvj (approved):

PR #1555, opened 2024-04-15
AUT-2393: Axios upgrade from 0.25.0 to 0.1.6.8

ayoshebby:
## What

Bumps [axios](https://github.com/axios/axios) from 0.25.0 to 1.6.8.

### Release notes:
**Bug Fixes**

- AxiosHeaders: fix AxiosHeaders conversion to an object during config merging (https://github.com/axios/axios/issues/6243) ([2656612](https://github.com/axios/axios/commit/2656612bc10fe2757e9832b708ed773ab340b5cb))
- import: use named export for EventEmitter; ([7320430](https://github.com/axios/axios/commit/7320430aef2e1ba2b89488a0eaf42681165498b1))
- vulnerability: update follow-redirects to 1.15.6 (https://github.com/axios/axios/issues/6300) ([8786e0f](https://github.com/axios/axios/commit/8786e0ff55a8c68d4ca989801ad26df924042e27))

### DIFF
--- a/package.json
+++ b/package.json
@@ -65,7 +65,7 @@
     "@otplib/core": "^12.0.1",
     "@otplib/plugin-base32-enc-dec": "^12.0.1",
     "aws-sdk": "^2.1356.0",
-    "axios": "^0.25.0",
+    "axios": "^1.6.8",
     "connect-redis": "^7.1.1",
     "cookie-parser": "^1.4.5",
     "copyfiles": "^2.4.1",
--- a/src/utils/http.ts
+++ b/src/utils/http.ts
@@ -10,7 +10,9 @@ import { ApiResponseResult } from "../types";
 import { HTTP_STATUS_CODES } from "../app.constants";
 import { ApiError } from "./error";
 
-const headers: AxiosRequestHeaders = {
+interface CustomAxiosRequestHeaders extends Partial<AxiosRequestHeaders> {}
+
+const headers: CustomAxiosRequestHeaders = {
   Accept: "application/json",
   "Content-Type": "application/json; charset=utf-8",
   "Access-Control-Allow-Credentials": "true",
--- a/src/utils/error.ts
+++ b/src/utils/error.ts
@@ -9,8 +9,8 @@ export class BadRequestError extends Error {
 
 export class ApiError extends Error {
   private status?: number;
-  private data?: string;
-  constructor(message: string, status?: number, data?: string) {
+  private data?: unknown;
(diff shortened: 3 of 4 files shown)


src/assets/javascript/cookies.js
```diff
@@ -70,7 +70,7 @@ var cookies = function (trackingId, analyticsCookieDomain) {
       setCookie(
         COOKIES_HISTORY_LENGTH,
         window.history.length
-        );
+      );
```
BeckaL, line 73: could we get rid of these unrelated changes from this commit?

src/utils/http.ts
```diff
@@ -97,7 +97,7 @@ export class Http {
       apiError = new ApiError(
         error.message,
         error.response.status,
-        error.response.data
+        error.response.data as string
```
BeckaL, line 100: Could you add some info into the commit message as to why we need this change? Is this an axios bump thing?

BeckaL (commented):
Could we reword the commit to remove the WIP comment? Also, what's our rollout plan, are we going to pause the pipeline?

package.json
```diff
@@ -65,7 +65,7 @@
     "@otplib/core": "^12.0.1",
     "@otplib/plugin-base32-enc-dec": "^12.0.1",
     "aws-sdk": "^2.1356.0",
-    "axios": "^0.25.0",
+    "axios": "^0.28.0",
```
mattvot, line 68: Is it possible to upgrade this to the latest `axios` version, currently `1.6.8`?

src/utils/http.ts
ayoshebby: There was a failure in the pre-merge checks regarding of error on that line:

`Argument of type 'unknown' is not assignable to parameter of type 'string'.
`
Only seems to come up when axios is upgraded. 

I've now bumped the version to 1.6.8, and that error no longer exists

package.json
ayoshebby: I've updated the version to 1.6.8 now, looks good asides from a change required in the https.ts file where it was enforcing to implement the full AxiosHeader properties

mattvot (commented):
Looking good. Just trying to assess if any more changes are required / we're not going to have hidden issues around. I've one suggestion for the moment.

mattvot (approved):
Just ran through some smoke testing and seen the flows work when targeting the build env. Approved.

mattvot (approved):

BeckaL (approved):

PR #1588, opened 2024-04-29
AUT-2733: Fix server sent consent cookie

gtvj:
## What

* Removes `HttpOnly` flag from `cookie_preferences_set` cookie included in HTTP responses. The [commit message](https://github.com/govuk-one-login/authentication-frontend/commit/b9e295d24fd67acfbd6d844fbd3030fba0334ee2) provides more context on the reasons for this change
* Removes duplicated `createConsentCookie` function that served only to wrap the `res.cookie` function provided by Express
* Applies `String` encoding to the cookie for consistency with encoding when the cookie is set on the client.
* Updates associated tests to ensure calls to `res.cookie` include the expected arguments and flags that will set the `HttpOnly` and `Secure` attributes. The [commit message](https://github.com/govuk-one-login/authentication-frontend/pull/1588/commits/939071fb1ceeaaa7c51132f76b1fdb7532d06e48) provides further information.
* Changes `CookieConsentModel`'s 'expiry' property to match the 'expires' property used in Express `res.cookie`

## How to review

1. Code Review
1. Run the tests
1. Run the code locally and, using an Incognito window (or having cleared cookies for the host), verify that:
   a. The cookie is set to reflect the HTTP headers included in the response from `/authorize`
   b. The server-sent cookie value is not URL encoded
   c. The banner is not shown 
   d. Interacting with the cookie banner updates the cookie value as expected (both when accepting and rejecting). You'll need to remove the server-sent cookie, reload the page, change settings and proceed to the next page to do this

## Related PRs

- https://github.com/govuk-one-login/authentication-frontend/pull/1224 introduced the `HttpOnly` flag for the `cookie_preferences_set` cookie

### DIFF
--- a/src/components/authorize/authorize-controller.ts
+++ b/src/components/authorize/authorize-controller.ts
@@ -38,7 +38,7 @@ function createConsentCookie(
   res.cookie(COOKIES_PREFERENCES_SET, consentCookieValue.value, {
     expires: consentCookieValue.expiry,
     secure: true,
-    httpOnly: true,
+    httpOnly: false,
     domain: res.locals.analyticsCookieDomain,
   });
 }
--- a/src/components/authorize/authorize-controller.ts
+++ b/src/components/authorize/authorize-controller.ts
@@ -152,7 +152,7 @@ export function authorizeGet(
         cookiesConsentService.createConsentCookieValue(cookieConsent);
 
       res.cookie(COOKIES_PREFERENCES_SET, consentCookieValue.value, {
-        expires: consentCookieValue.expiry,
+        expires: consentCookieValue.expires,
         secure: true,
         httpOnly: false,
         domain: res.locals.analyticsCookieDomain,
--- a/src/components/authorize/authorize-controller.ts
+++ b/src/components/authorize/authorize-controller.ts
@@ -30,7 +30,7 @@ import { Claims } from "./claims-config";
 
 export function authorizeGet(
   authService: AuthorizeServiceInterface = authorizeService(),
-  cookieService: CookieConsentServiceInterface = cookieConsentService(),
+  cookiesConsentService: CookieConsentServiceInterface = cookieConsentService(),
   kmsService: KmsDecryptionServiceInterface = new KmsDecryptionService(),
   jwtService: JwtServiceInterface = new JwtService()
 ): ExpressRouteFunc {
@@ -149,7 +149,7 @@ export function authorizeGet(
 
     if (req.session.client.cookieConsentEnabled && cookieConsent) {
       const consentCookieValue =
-        cookieService.createConsentCookieValue(cookieConsent);
+        cookiesConsentService.createConsentCookieValue(cookieConsent);
 
       res.cookie(COOKIES_PREFERENCES_SET, consentCookieValue.value, {
         expires: consentCookieValue.expiry,
--- a/src/components/authorize/authorize-controller.ts
+++ b/src/components/authorize/authorize-controller.ts
@@ -8,10 +8,7 @@ import {
 import { getNextPathAndUpdateJourney, ERROR_CODES } from "../common/constants";
 import { BadRequestError, QueryParamsError } from "../../utils/error";
 import { ExpressRouteFunc } from "../../types";
-import {
-  CookieConsentModel,
-  CookieConsentServiceInterface,
-} from "../common/cookie-consent/types";
+import { CookieConsentServiceInterface } from "../common/cookie-consent/types";
 import { cookieConsentService } from "../common/cookie-consent/cookie-consent-service";
 import { sanitize } from "../../utils/strings";
 import { USER_JOURNEY_EVENTS } from "../common/state-machine/state-machine";
@@ -30,18 +27,6 @@ import {
 } from "../../config";
 import { logger } from "../../utils/logger";
 import { Claims } from "./claims-config";
-
-function createConsentCookie(
-  res: Response,
-  consentCookieValue: CookieConsentModel
-) {
-  res.cookie(COOKIES_PREFERENCES_SET, consentCookieValue.value, {
-    expires: consentCookieValue.expiry,
-    secure: true,
-    httpOnly: false,
-    domain: res.locals.analyticsCookieDomain,
-  });
-}
 
 export function authorizeGet(
   authService: AuthorizeServiceInterface = authorizeService(),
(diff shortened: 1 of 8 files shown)


src/components/authorize/authorize-controller.ts
```diff
@@ -166,7 +151,13 @@ export function authorizeGet(
...
         cookieService.createConsentCookieValue(cookieConsent);
 
-      createConsentCookie(res, consentCookieValue);
+      res.cookie(COOKIES_PREFERENCES_SET, consentCookieValue.value, {
+        expires: consentCookieValue.expiry,
+        secure: true,
+        httpOnly: false,
+        domain: res.locals.analyticsCookieDomain,
+        encode: String,
+      });
```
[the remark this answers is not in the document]
gtvj: Hi @andrew-moores - I've introduced an additional commit that extends the tests to assert that calls to `res.cookie` include the expected arguments. I initially had a bit of difficulty because the `expiry` property was changing between calls (by a second) and causing failures, but found I could assert for only the `httpOnly`, `secure` and `encode` properties by passing these as keys to an object passed to `sinon.match`

src/components/common/cookies/cookies-controller.ts
```diff
@@ -27,22 +26,16 @@ export function cookiesPost(req: Request, res: Response): void {
...
-  res: Response,
-  consentCookieValue: CookieConsentModel
-) {
   res.cookie(COOKIES_PREFERENCES_SET, consentCookieValue.value, {
     expires: consentCookieValue.expiry,
     secure: true,
-    httpOnly: true,
+    httpOnly: false,
     domain: res.locals.analyticsCookieDomain,
+    encode: String,
```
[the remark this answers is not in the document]
gtvj: Hi @andrew-moores - I've introduced an additional commit that extends the tests to assert that calls to `res.cookie` include the expected arguments. I initially had a bit of difficulty because the `expiry` property was changing between calls (by a second) and causing failures, but found I could assert for only the `httpOnly`, `secure` and `encode` properties by passing these as keys to an object passed to `sinon.match`

src/components/authorize/tests/authorize-controller.test.ts
```diff
@@ -108,7 +111,15 @@ describe("authorize controller", () => {
         fakeJwtService
       )(req as Request, res as Response);
 
-      expect(res.cookie).to.have.been.called;
+      expect(res.cookie).to.have.been.calledWith(
+        COOKIES_PREFERENCES_SET,
+        consentCookieValue.value,
+        sinon.match({
```
andrew-moores, line 117:
```suggestion
        sinon.match({
        expires: sinon.match((date: Date) => {
            const expectedExpires = new Date(Date.now());
            expectedExpires.setFullYear(expectedExpires.getFullYear() + 1)
            return Math.abs(date.getTime() - expectedExpires.getTime()) < 1000;
          }),
```

test/helpers/mock-cookie-consent-service-helper.ts
```diff
@@ -0,0 +1,20 @@
+import { sinon } from "../utils/test-utils";
+import { CookieConsentServiceInterface } from "../../src/components/common/cookie-consent/types";
+
+export function createMockCookieConsentService(
+  userCookieConsentPreference: string
+): CookieConsentServiceInterface {
```
andrew-moores, line 6:
```suggestion
): CookieConsentServiceInterface {
  const expiryDate: Date = new Date();
  expiryDate.setFullYear(expiryDate.getFullYear() + 1);

```

test/helpers/mock-cookie-consent-service-helper.ts
```diff
@@ -0,0 +1,20 @@
...
+    getCookieConsent: sinon.fake(),
+    createConsentCookieValue: sinon.fake.returns({
+      value: JSON.stringify({
+        analytics: userCookieConsentPreference === "true",
+      }),
+      expires: (function (): any {
+        const cookieExpires = new Date();
+        cookieExpires.setFullYear(cookieExpires.getFullYear() + 1);
+        return cookieExpires;
+      })(),
```
andrew-moores, line 17:
```suggestion
    expiry: expiryDate
```

src/components/authorize/authorize-controller.ts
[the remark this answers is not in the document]
andrew-moores: I've added some suggestions for testing the expiry date as it seems like an important feature to check.

src/components/authorize/tests/authorize-controller.test.ts
andrew-moores: Check that the expiry date is approximately one year in the future from now allowing a difference of 1000 milliseconds to account for the time the test takes to run.

test/helpers/mock-cookie-consent-service-helper.ts
andrew-moores, line 1: Ideally we should use the actual service rather than a mock here as the expiry date is different depending on the consent given by the user.

test/helpers/mock-cookie-consent-service-helper.ts
andrew-moores: The `CookieConsentModel` uses expiry but the express Request cookie method uses expires, ideally the model should be updated to align with express.

test/helpers/mock-cookie-consent-service-helper.ts
gtvj: Thanks @andrew-moores - as discussed, I've got the go-ahead to create a technical debt ticket for this change and created it in Jira so it can be triaged/prioritised and brought in to a future sprint. I've shared details of the ticket on Slack.

src/components/authorize/authorize-controller.ts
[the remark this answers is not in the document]
gtvj: Thanks @andrew-moores - I've included all the suggestions now.

test/helpers/mock-cookie-consent-service-helper.ts
gtvj: Hi @andrew-moores - thanks for this suggestion. I've just pushed [a new commit](https://github.com/govuk-one-login/authentication-frontend/pull/1588/commits/813241569875bb7b13dae819a7ec33ea4b87ff9c) that does this refactoring

andrew-moores (approved):

BeckaL (approved):
Approving on Andrew's behalf

PR #1625, opened 2024-05-16
Aut 2578/use common headers

BeckaL:
## Checklist


src/components/check-reauth-users/tests/check-reauth-user-service.test.ts
```diff
@@ -1,22 +1,66 @@
...
+        email,
+        SUBJECT,
+        ip,
+        clientSessionId,
+        diPersistentSessionId
+      );
+
+    const expectedApiCallDetails = {
+      expectedPath: API_ENDPOINTS.CHECK_REAUTH_USER,
+      expectedHeaders: expectedHeadersFromCommonVarsWithoutSecurityHeaders,
```
andrew-moores, line 55: Why do we need to pass this?  Maybe it could be a way to add additional headers to be checked rather simply passing a constant from the service-test-helper module back to it.

PR #1626, opened 2024-05-17
get Dev image for Dev frontend deployment

pskushwaha1:
## What

Updating the get-image script so for DEV frontend deployment it pull image from DEV tooling frontend Repo 

## How to review

1. Code Review

### DIFF
--- a/ci/terraform/get_image.sh
+++ b/ci/terraform/get_image.sh
@@ -2,17 +2,27 @@
 
 set -euo pipefail
 
-GITHUB_SHA=$1
-REGISTRY_ID=114407264696
+GITHUB_SHA="${1}"
+ENVIRONMENT="${2}"
 FRONTEND_REPO_NAME=frontend-image-repository
 SIDECAR_REPO_NAME=basic-auth-sidecar-image-repository
+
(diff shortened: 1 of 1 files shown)


ci/terraform/get_image.sh
```diff
@@ -3,10 +3,22 @@
...
+account_id=$(aws sts get-caller-identity|jq -r ".Account")
+
+echo "Setting the ECR repo registry ID"
+# Check for Auth DEV Account ID  if true then pull image from tools-dev Dev to deploy Dev frontend 
+# Else pull image from tools-prod to deploy  ( Build , integration , Staging & prod  ) frontend 
+if [ "$account_id" = 653994557586 ]; then 
+  REGISTRY_ID=706615647326
+  else 
+  REGISTRY_ID=114407264696
+fi
```
whi-tw, line 20: I'm really not a huge fan of this approach - is there a way we can make this selection without hard-coding the account IDs in here? I'd generally rather not expose the account IDs directly.

Would there be a way to do this via secrets / an argument into the script?

whi-tw (changes requested):

whi-tw (approved):
LGTM!

PR #1640, opened 2024-05-22
Bau: Remove ip from service calls

BeckaL:
## What

Removes sourceIp as a parameter from all remaining services (some have already been done as part of AUT-2578).

This was previously being used to set the "x-forwarded-for" header on the requests we send to the backend, but the common headers library is now being used to set this header, which requires the whole request object to be passed through, which we have already done [here](https://github.com/govuk-one-login/authentication-frontend/pull/1635).

Therefore these services can be tidied by removing this parameter, and also removing it from the overall options.

This PR also contains some minor tidies that I came across along the way.

## How to review


gtvj (approved):
Hi @BeckaL - I've been through the PR commit by commit and all looks good.

PR #1674, opened 2024-06-11
AUT-2729: Fix text wrapping issue on home page

LazarAlexandru-Constantin:
## What

Add a non-breaking space between the words "sign" and "in".  This pr is a 1 to 1 copy of [this PR](https://github.com/govuk-one-login/authentication-frontend/pull/1564), that was reverted.

##Why

Fixes an issue raised by UCD where the word "in" had been appearing on a line alone following a recent content change. This has been fixed by introducing non-breaking space HTML entities to ensure the words in 'sign in' and 'One Login' do not split across lines as text wraps across different screen sizes.

## How to review

1. Code Review

### DIFF
--- a/src/components/sign-in-or-create/index.njk
+++ b/src/components/sign-in-or-create/index.njk
@@ -17,7 +17,7 @@
 {% block content %}
     {% include "common/errors/errorSummary.njk" %}
 
-    <h1 class="govuk-heading-l govuk-!-margin-top-0 govuk-!-margin-bottom-3">{{ 'pages.signInOrCreate.mandatory.header' | translate }}</h1>
+    <h1 class="govuk-heading-l govuk-!-margin-top-0 govuk-!-margin-bottom-3">{{ 'pages.signInOrCreate.mandatory.header' | translate | striptags | safe }} </h1>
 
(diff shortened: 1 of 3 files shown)


gtvj (approved):

src/components/sign-in-or-create/index.njk
```diff
@@ -17,7 +17,7 @@
 {% block content %}
     {% include "common/errors/errorSummary.njk" %}
 
-    <h1 class="govuk-heading-l govuk-!-margin-top-0 govuk-!-margin-bottom-3">{{ 'pages.signInOrCreate.mandatory.header' | translate }}</h1>
+    <h1 class="govuk-heading-l govuk-!-margin-top-0 govuk-!-margin-bottom-3">{{ 'pages.signInOrCreate.mandatory.header' | translate | striptags | safe }} </h1>
```
dbes-gds, line 20: this is applied to the header rather than the title, was that the intent?

LazarAlexandru-Constantin: Yes, the header is the intent, the title should remain unchanged. The non breaking space should be in the header so the 2 words go together. There is no need for a &nbsp in the title

PR #1677, opened 2024-06-11
BAU: Add a forwarded_ip_config block to rate rule

whi-tw:
## What

This will allow the WAF to use the X-Forwarded-For header in the case that it's provided (request via cloudfront). It will fall back to using the 'true' client IP if the header is not provided (direct request).

## How to review

Code review


galund:
This is not the mitigation previously agreed to let CloudFront traffic through, and doesn't meet the security requirements.

PR #1783, opened 2024-07-04
BAU:fix Sonarcloud "Expected the Promise rejection reason to be an Er…

LazarAlexandru-Constantin:
…ror."

Using an instance of the Error class allows you to provide more meaningful information about the error. The Error class and its subclasses provide properties such as message and stack that can be used to convey useful details.

## What

Fix [this](https://sonarcloud.io/project/issues?resolved=false&sinceLeakPeriod=true&types=CODE_SMELL&id=govuk-one-login_authentication-frontend&open=AY8Ub9V0CJ2aVgC8DLNx&tab=code) Sonarcloud issue.

## How to review

1. Code Review

### DIFF
--- a/src/components/common/constants.ts
+++ b/src/components/common/constants.ts
@@ -190,7 +190,7 @@ export async function getNextPathAndUpdateJourney(
   await new Promise<void>((resolve, reject) => {
     req.session.save((error) => {
       if (error) {
-        reject(error);
+        reject(new Error(error));
         req.log.error(
           "Session could not be saved after setting the user journey."
         );


alhcomer (approved):

mattvot (commented):
Hey @LazarAlexandru-Constantin 👋 do you know what the type of `error` is. It may be that `express-session` is already returning an Error, but Typescript isn't aware of it.

LazarAlexandru-Constantin:
Hello @mattvot!
I looked at the documentation and and the return type of the error object is `any`. 
Indeed I think the case is that `express-session` is returning an `Error` object. 
Do you think this implementation is breaking or should be reverted?

PR #1785, opened 2024-07-04
BAU: add metadata that codepipeline understands

whi-tw:
## What

As-per https://ordinaryexperts.com/blog/2019/04/27/better-codepipeline-metadata

This extra metadata is shown in the codepipeline UI and will allow better identification of where the artifact has come from.

image

## How to review

- Code review
- Deploy to dev from this branch (`./scripts/trigger-gha-current-branch.sh`) and confirm that the resulting pipeline run has the sensible metadata

Unfortunately the pr merge one is untestable without merging a pr. I guess 'trust me'?

## Related PRs

- govuk-one-login/authentication-api#4851

### DIFF
--- a/.github/workflows/build-deploy-frontend-dev.yml
+++ b/.github/workflows/build-deploy-frontend-dev.yml
@@ -12,6 +12,102 @@ on:
   workflow_dispatch:
 
 jobs:
+  pr-data:
+    name: Get data for merged PR
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      pull-requests: read
+    outputs:
+      data: ${{ steps.get_pr_data.outputs.result }}
+    steps:
+      - name: Get PR data
+        uses: actions/github-script@v7
+        id: get_pr_data
+        with:
+          script: |
+            const query = `query($owner: String!, $name: String!, $oid: GitObjectID!) {
+                repository(owner: $owner, name: $name) {
+                  object(oid: $oid) {
+                    ... on Commit {
+                      oid
+                      message
+                      associatedPullRequests(first: 1) {
+                        nodes {
+                          number
+                          title
+                          merged
+                          mergedAt
+                          mergeCommit {
+                            oid
+                          }
+                        }
+                      }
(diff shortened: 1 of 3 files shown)


.github/workflows/build-deploy-frontend-dev.yml
```diff
@@ -87,6 +185,6 @@ jobs:
             --bucket ${{ env.DEV_ARTIFACT_BUCKET }} \
             --key frontend.zip \
             --body frontend.zip \
-            --metadata "repository=$GITHUB_REPOSITORY,commitsha=$GITHUB_SHA,committag=$GIT_TAG,commitmessage=$COMMIT_MSG"`
+            --metadata '${{ toJson(fromJson(needs.pr-data.outputs.data)) }}'`
```
mattvot, line 189: Is the not json => json conversion here just for condensing the string length?

```suggestion
            --metadata '${{ needs.pr-data.outputs.data }}'`
```

mattvot (approved):
Looks great, thanks! Can see this being a big improvement to our pipeline visibility. 

Two questions but good to go.

Is it possible/convenient to dedupe the JS code here? Totally appreciate we've already got some duplication on these pipelines.

.github/workflows/build-deploy-frontend-dev.yml
whi-tw: ref. dedupe: splitting the js out to a separate file is a nightmare with github script - you're welcome to give it a go, though!.

I guess we could split this job out to a separate reusable workflow, but the scope access can get a bit weird. I think a full dedupe is necessary on these workflows - they are on my hit list for future.

ref. json -> json, agreed, it looks janky as hell. the issue is that the output from the response isn't 'proper' json. The script action is weird - it returns a javascript object, which is 'kinda' json'y, but doesn't properly parse when passed on to aws. this to/from shuffle sorts it out and presents it to AWS in a way it can cope with. I did try it without, but it's super flaky

mattvot (approved):

PR #1795, opened 2024-07-08
BAU: Simplify frontend deployment process

whi-tw:
- **BAU: Update pre-commit hooks**
- **BAU: Merge build and deploy frontend workflows**

## What

Previously we had two workflows:

1. Build and push the docker image
2. Create and upload the terraform zip

The second workflow was triggered by the completion of the first.

This could lead to a situation where due to network slowdowns, two PRs merged in a short space of time could have their metadata (and possibly terraform) mixed up: the checkout stage in the deploy job just pulled the HEAD commit from the repo, regardless of the commit the `build and push` job was spawned from.

In this new single workflow, we build the image and push the terraform within the same workflow run, meaning that race conditions are no longer possible.

Additionally, I've added a concurrency group, to ensure that only one deploy job can run at once.

Finally, I've added an actionlint pre-commit hook and improved the two `node` hooks - instead of building up a docker container each time, it just uses the local `node_modules` folder and runs the tasks natively. This is far faster.

## Checklist


BeckaL (approved):
Looks sensible to me

PR #1804, opened 2024-07-09
LS-2142 Fix ScaleDown

andyloughran:
## What

Autoscaling policy was not scaling down, as CPU for stepadjustment was being set to 0

## How to review

See discussion in comments.


andyloughran:
image

pskushwaha1 (approved):

PR #1805, opened 2024-07-09
AUT-2790: Add logic to redirect to sign-out when introducing wrong email is introduced in reauth

VladGavrilet:
## What
 AUT-2790: Add logic to redirect to sign-out when introducing wrong email in re-auth journey

## How to review

1. Code Review
2. Deploy to authdev1, with [this](https://github.com/govuk-one-login/authentication-api/pull/4899) as back-end and test that you are logged out, after 5 attempts with a wrong email in the re-auth journey
3. Try to log in and check that you are not locked

## Checklist

- [x] Performance analyst has been notified of the change.

## Related PRs
https://github.com/govuk-one-login/authentication-api/pull/4899

### DIFF
--- a/src/components/enter-email/tests/enter-email-integration.test.ts
+++ b/src/components/enter-email/tests/enter-email-integration.test.ts
@@ -1,5 +1,5 @@
 import request from "supertest";
-import { describe } from "mocha";
+import { afterEach, describe } from "mocha";
 import { expect, sinon } from "../../../../test/utils/test-utils";
 import * as cheerio from "cheerio";
 import decache from "decache";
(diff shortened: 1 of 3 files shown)


src/components/common/constants.ts
LazarAlexandru-Constantin, line 1: verify if this is still needed

di-fabs (approved):

src/components/enter-email/tests/enter-email-integration.test.ts
```diff
@@ -8,23 +8,50 @@ import {
...
 
 describe("Integration::enter email", () => {
   let token: string | string[];
   let cookies: string;
   let app: any;
   let baseApi: string;
+  let checkReauthUsersService: any;
+  let stubConfig: { statusCode: number; shouldStub: boolean };
+
+  function setupCheckReauthUsersStub() {
```
BeckaL, line 28: I'm not sure this function makes sense for a couple of reasons:

Given it's an integration test, do we instead want to be testing the integration with the reauth users service, and using nock to capture the http call?

If not, this function as is looks like it could be simplified in a few ways:

* add the relevant params to the function so that we're not reading from external state - I can't see any reason here why we need the stub config set as a let outside of this
* When would we not want to stub (`shouldStub` = false?) To me that branch doesn't really make any sense

src/components/enter-email/tests/enter-email-integration.test.ts
```diff
@@ -247,4 +281,38 @@ describe("Integration::enter email", () => {
...
   });
+
+  it("should redirect to /signed-out with login_required error when user fails re-auth", async () => {
+    process.env.SUPPORT_REAUTHENTICATION = "1";
+    stubConfig = {
+      statusCode: HTTP_STATUS_CODES.BAD_REQUEST,
+      shouldStub: true,
+    };
+
+    nock(baseApi)
```
BeckaL, line 292: e.g. here we've got `nock` (which I think is the way to go), but also an unused `stubConfig` above

src/components/enter-email/tests/enter-email-integration.test.ts
andrew-moores: I've removed this as its not used.

BeckaL (approved):

PR #1822, opened 2024-07-11
AUT-2789: Log user out when they enter their password incorrectly max…

andrew-moores:
… allowed times.

Modifying the integration tests so they wait for the initial call to the app executed in the before method to complete.  This call is needed so the tests can use the csrf token and the correct cookies so it must complete before any of the tests are executed.

## What

Some of the integration tests have been updated so they work correctly by enforcing a wait on the initial call to the service to obtain the tokens required in the subsequent tests.

The changes to support the reauth journey are dependent on matching changes in the backend that signal the user should be logged out rather than locked out if they fail to enter their password correctly more than 5 times.

When the back end signal the user has failed to provide their password correctly after 5 attempts the frontend checks if the journey is a reauth journey and then logs the user out by making a call to orchestration with the new path parameter  `error` set to `login_required`. 

## How to review
1. Review backend changes for context.
2. Code review
3. Deploy to environment (or run locally) and walk through the reauth flow while entering an invalid password.

## Checklist

### DIFF
--- a/src/app.ts
+++ b/src/app.ts
@@ -169,7 +169,7 @@ async function createApp(): Promise<express.Application> {
   app.use(setLocalVarsMiddleware);
   app.use(setGTM);
 
-  i18next
+  await i18next
     .use(Backend)
     .use(i18nextMiddleware.LanguageDetector)
     .init(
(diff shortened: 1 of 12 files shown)


BeckaL (commented):
Looks good to me - just a quick check before I approve: Do we need to feature flag the code changes (or e.g. will the reauth journey type only be set on the session when the feature flag is on)? Just a check on my side as I haven't been as embroiled in the code or feature flags discussions as you.

andrew-moores:
> Looks good to me - just a quick check before I approve: Do we need to feature flag the code changes (or e.g. will the reauth journey type only be set on the session when the feature flag is on)? Just a check on my side as I haven't been as embroiled in the code or feature flags discussions as you.

@BeckaL When the reauth feature switch is off in production if an RP attempts to send the user for reauth then the reauth pages will not be available and the user will simply be returned to the RP as logged in without re-entering any credentials.  When the feature switch is on the user will be presented with the reauth flow pages.  If they fail to present one of their credentials too many times during this flow they will be logged out.

BeckaL (approved):

PR #1863, opened 2024-07-22
AUT-1723: Trust all proxies

whi-tw:
## What

As a) this is only accessible via cloudfront and b) cloudfront will always set the CloudFront-Viewer-Address header, we can trust all proxies - it won't be possible for someone to forge the header.

## How to review

1. code review
2. ensure tests have passed.


basic-auth-sidecar/nginx.conf
```diff
@@ -20,7 +20,8 @@ http {
 
     log_format  main  '$remote_addr - $remote_user [$time_local] "$request" '
                       '$status $body_bytes_sent "$http_referer" '
-                      '"$http_user_agent" "$http_x_forwarded_for"';
+                      '"$http_user_agent" "$http_x_forwarded_for"'
```
mattvot, line 23: should this log line have a white space character at the end like the above?
```suggestion
                      '"$http_user_agent" "$http_x_forwarded_for" '
```

whi-tw: No, not this one. These lines are concatenated as-is, ie.

```
'"hello" "world" '
'"my" "name" "is" "Tom"'

=>

'"hello" "world" "my" "name" "is" "Tom"'
```

If the last line had a trailing space, the log line would be output with a trailing space:

```
'"hello" "world" '
'"my" "name" "is" "Tom" '

=>

'"hello" "world" "my" "name" "is" "Tom" '
```

Which is undesirable!

mattvot: Is that still true considering line 24 adds `'"http_cloudfront_viewer_address"'` to the end?

whi-tw: 🤦🏻‍♂️ ☠️

PR #1930, opened 2024-08-08
AUT-1555: Add journey to reset password routes

BeckaL:
Without this, a user was allowed to manually transition to these pages from an invalid state (e.g. their account was not found). This meant that they would end up hitting a backend error at some point, rather than this not being allowed in the first place.

Also added session middleware as it seems correct to be doing both of these checks

## How to review

1. Code Review
1. Run the app locally
1. When signing in, enter an email that doesn't have an account associated with it, and see that you hit the account not found error
1. Manually change the url to /reset-password-request, or /reset-password-required
1. See that you are redirected back to /account not found, and do not hit any unexpected errors
1. Compare this behaviour against staging
1. Also try a normal reset password journey locally, and verify that it still works.

### DIFF
--- a/src/components/reset-password/reset-password-routes.ts
+++ b/src/components/reset-password/reset-password-routes.ts
@@ -15,7 +15,12 @@ import { accountInterventionsMiddleware } from "../../middleware/account-interve
 
 const router = express.Router();
 
-router.get(PATH_NAMES.RESET_PASSWORD_REQUEST, resetPasswordRequestGet);
+router.get(
+  PATH_NAMES.RESET_PASSWORD_REQUEST,
+  validateSessionMiddleware,
+  allowUserJourneyMiddleware,
+  resetPasswordRequestGet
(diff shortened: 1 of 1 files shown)


gtvj:
Hi @BeckaL, I've reviewed the code and manually checked steps 2 to 7. All looks good and working OK. 

I did hit a snag when trying a normal password reset journey locally because I'm getting directed to an unstyled page on the build stub on path `/oidc/authorization-code/callback` 

> Error in Callback
> Error: server_error
> Error description: Unexpected server error 

I'm also getting this when trying to create an account locally.

I'm confident this isn't related to this PR though because I'm seeing the same behaviour on `main`.

gtvj:
Thanks for the suggestion about updating my environment variables. Having done so I'm seeing the expected page at the end of the password reset journey locally.

gtvj (approved):

PR #1109, opened 2023-07-20
FIX: Prevent account recovery link display where not permitted

gtvj:
## What?

Amends Nunjucks templates for entering Auth App and SMS codes during sign in. The change strengthens conditional to require strict equality. Also introduces some associated integration tests.

## Why?

Bug identified where the a non-functional "You can securely change how you get security codes" link is presented to users (if users click on this link they are presented with the "Sorry, there is a problem" page).

PR #1931, opened 2024-08-08
AUT-3052: Display Link on MFA code entery page

VladGavrilet:
## What

Ensured the details section with the 'I do not have access to the authenticator app' link remains visible when an incorrect authenticator app code is entered.

## How to review

1. Code Review
2. Deploy to a dev environment and test:
-Enter correct email address, press Continue.
-Enter correct password, press Continue.
-The ‘Enter the 6 digit security code shown in your authenticator app’ page is displayed with the ‘I do not have access to the authenticator app’ link.
-Enter an incorrect auth app code, press Continue. The error message is displayed for the incorrect code, but the ‘I do not have access to the authenticator app’ link is remains visible.

### DIFF
--- a/src/components/enter-authenticator-app-code/index.njk
+++ b/src/components/enter-authenticator-app-code/index.njk
@@ -40,7 +40,6 @@
 
   </form>
 
-  {% if isAccountRecoveryPermitted === true %}
     {% set detailsHTML %}
       <p class="govuk-body">
         {{'pages.enterAuthenticatorAppCode.details.text1' | translate}}
@@ -52,7 +51,6 @@
       summaryText: 'pages.enterAuthenticatorAppCode.details.summary' | translate,
       html: detailsHTML
     }) }}
-  {% endif %}
 
   {{ga4OnPageLoad({ nonce: scriptNonce, statusCode: "200", englishPageTitle: pageTitleName, taxonomyLevel1: "authentication", taxonomyLevel2: "sign in", contentId: "89461417-df3f-46a8-9c37-713b9dd78085", loggedInStatus: false, dynamic: false })}}
 {% endblock %}
--- a/src/components/enter-authenticator-app-code/index.njk
+++ b/src/components/enter-authenticator-app-code/index.njk
@@ -12,7 +12,7 @@
   <form id="form-tracking" action="/enter-authenticator-app-code" method="post" novalidate="novalidate">
 
     <input type="hidden" name="_csrf" value="{{csrfToken}}"/>
-    <input type="hidden" name="isAccountRecoveryPermitted" value="{{isAccountRecoveryPermitted}}"/>
+    <input type="hidden" name="isAccountRecoveryPermitted" value="{{isAccountRecoveryPermitted | default(false)}}"/>
     <input type="hidden" name="checkEmailLink" value="{{checkEmailLink}}"/>
 
     {{ govukInput({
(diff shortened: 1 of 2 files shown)


src/components/enter-authenticator-app-code/index.njk
```diff
@@ -40,7 +40,6 @@
 
   </form>
 
-  {% if isAccountRecoveryPermitted === true %}
```
BeckaL, line 43: Hmm, I'm not sure about this. Presumably there's a reason for this line outside of the case you're looking at. Are we sure this can be totally discarded (as opposed to tweaking something for the thing we're trying to fix)

VladGavrilet: From what I managed to understand, this should be ok. When the user is required a higher level of trust another template is being rendered "index-2fa-service-uplift-auth-app.njk". There the link will still not be visible after a wrong code is submitted in those cases.

BeckaL: Cool, I'll let others wade in on this (I see @gtvj did some of these changes). But this kind of context is really important to put in commit messages if you've found out why something that looks like it's needed isn't needed

BeckaL: Some of the smallest changes have the most context to them - see https://dhwthompson.com/2019/my-favourite-git-commit for a fun example

gtvj: Hello @VladGavrilet @BeckaL 👋🏽 

I've just been looking back through the code and it seems the original condition was introduced in [the PR](https://github.com/govuk-one-login/authentication-frontend/pull/964) for AUT-1088 to link the availability of the `<details>` element to account recovery being enabled for “the environment and the user”. I can see from that initial PR that:

* both `req.session.user.isAccountRecoveryPermitted` and `isAccountRecoveryPermittedForUser` are assigned to what we get back from the `accountRecoveryResponse` (See [here](https://github.com/govuk-one-login/authentication-frontend/blob/e0174382775969c4a09e0600c627538f0ddd4898/src/components/enter-authenticator-app-code/enter-authenticator-app-code-controller.ts#L80))
* `isAccountRecoveryPermittedForUser ` is what we pass through to the template as `isAccountRecoveryPermitted` (See [here](https://github.com/govuk-one-login/authentication-frontend/blob/e0174382775969c4a09e0600c627538f0ddd4898/src/components/enter-authenticator-app-code/enter-authenticator-app-code-controller.ts#L91)) 
* the API handler can return false for `accountRecoveryPermitted` (See [here](https://github.com/govuk-one-login/authentication-api/blob/bf8b6733f5fdef4d5724f1ec5e4596fd3e957864/frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/AccountRecoveryHandler.java#L95) - sorry in advance if I’ve misunderstood this, I'm not the best at interpreting Java code)

If that’s the case, it does seem there is a case where the `{% if isAccountRecoveryPermitted === true %}` condition would still need to apply, in which case I think we might need a different approach. 

I actually wonder if the problem here is that `accountRecoveryPermitted` isn’t being carried through to the template following a validation error (have a vague memory that we might have seen this kind of thing before). 

Sorry this took me a while to interpret - I think my only involvement in this before was introducing a [strict equality check](https://github.com/govuk-one-login/authentication-frontend/pull/1109). Looking at that PR the bug I was fixing arose from users being presented with the link when it shouldn't have been available to them. 

I feel like it might be good to include some of this context in the updated PR and commit message.

PR #1950, opened 2024-08-14
BAU: Fix shutdown script

alhcomer:
## What

Changed 'docker-compose' to 'docker compose' in the shutdown.sh script as 'docker-compose' has been depracated:

"GitHub deprecated v1, and you need to change the command from, e.g., docker-compose build to docker compose build (remove the dash)" - mentioned [here](https://github.com/orgs/community/discussions/116610)

## How to review

1. Code Review


BeckaL (approved):
So I'd just amend the PR wording here - in my understanding it's not that docker-compose is an invalid command, it's just a separate binary that some people don't have installed by default.

PR #1952, opened 2024-08-14
Aut 1541/playback partially obscured email

BeckaL:
## What

On uplift journey, we'd like to display a partially obscured email to users so they know which account they're signed in with.

This functionality has been feature flagged, to allow us to release to various environments separately. (Currently on switched on in sandpit and authdevs).

I have chosen not to introduce a dependency for the email obscuring, as the obscuring is simple and I'd rather not introduce a separate dependency just for this.

If there are problems obscuring the email (it's not present, or it's invalid), we log an error and do not display anything to the user. This should not happen in practice, and should only result from a programming error, hence this error handling strategy.

Do not merge yet - needs review by UCD and security.

## How to review

For example:

1. Code Review
1. Run locally with the env variable "SHOW_REDACTED_EMAIL_ON_UPLIFT_ENABLED" set to "1", and run through an uplift journey. I actually did this by hardcoding the value of "req.session.user.upliftJourney" in the relevant controllers and running through a normal sign in journey, since I struggled to do normal uplift journey locally.
1. See that you can see an obscured email on this screen, according to the rules in the ticket.
1. Repeat for the other auth type (SMS or auth app)
1. Set the feature flag to off, and observe that you don't see this screen.

## Checklist

<!-- UCD review


src/components/enter-authenticator-app-code/enter-authenticator-app-code-controller.ts
```diff
@@ -55,9 +59,16 @@ export function enterAuthenticatorAppCodeGet(
...
     if (!isAccountRecoveryEnabledForEnvironment) {
-      return res.render(templateName, {
+      let templateOptions: object = {
         isAccountRecoveryPermitted: false,
-      });
+      };
+      if (showRedactedEmail) {
+        templateOptions = {
+          ...templateOptions,
+          redactedEmail: redactEmail(email),
```
mattvot, line 68: Not a huge thing and I'm happy to leave as is. You may be able to save a little verbosity with a ternary here, assuming the template parser understands the undefined variable

e.g. 
```
return res.render(templateName, {
  isAccountRecoveryPermitted: false,
  redactedEmail: showRedactedEmail ? redactEmail(email) : undefined,
});
```

mattvot (approved):
Nice, I'm happy with the implementation, with the caveats regarding domain redaction as discussed.

+1 for no new dependency!

PR #2006, opened 2024-08-30
ATO-984 (2/5): Pass `previousSessionId` to StartHandler in POST request

billrdunn:
## Background

`previousSessionId` will be added as a claim in authorize request to Authentication (see 3 below).

This is required in order to break session dependencies between Authentication and Orchestration. When Auth has its own session store (see 2 below), the hash key for the session will be `sessionId`. When `sessionId` gets updated, the record will need to be recreated, and in order to do this, Auth need `previousSessionId` in order to map the previous session ID to the new one (which will be extracted from the gs cookie).

## What

Originally `previousSessionId` was added as a header in the GET request to /start. This has since been changed due to [this comment](https://github.com/govuk-one-login/authentication-api/pull/5154#discussion_r1738887805), but is still present in the commit history.

Request to /start has been changed from GET to POST. This allows `previousSessionId` (and upcoming data needed to passed via API to split Auth / Orch session dependency) to be sent in the request body.

## Related PRs

ATO-984 incremental deployment:
1. https://github.com/govuk-one-login/authentication-api/pull/5171
2. https://github.com/govuk-one-login/authentication-frontend/pull/2006
3. https://github.com/govuk-one-login/authentication-api/pull/5172
4. https://github.com/govuk-one-login/authentication-api/pull/5174
5. TODO: Remove GET from /start methods

### DIFF
--- a/src/components/authorize/tests/authorize-service.test.ts
+++ b/src/components/authorize/tests/authorize-service.test.ts
@@ -15,7 +15,7 @@ import {
 } from "../../../../test/helpers/service-test-helper";
 import { commonVariables } from "../../../../test/helpers/common-test-variables";
 
-const oldSessionId = "4waJ14KA9IyxKzY7bIGIA3hUDos";
+const previousSessionId = "4waJ14KA9IyxKzY7bIGIA3hUDos";
 
 describe("authorize service", () => {
   let postStub: SinonStub;
@@ -102,7 +102,7 @@ describe("authorize service", () => {
     ).to.be.true;
   });
 
-  it("sends a request with old session ID in the body when given", () => {
+  it("sends a request with previous session ID in the body when given", () => {
     process.env.SUPPORT_REAUTHENTICATION = "1";
     service.start(
       sessionId,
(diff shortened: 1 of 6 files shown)


mrwilson:
Two questions - neither of which are blockers:
- How do you feel about `previousSessionId` rather than `oldSessionId`? Is there a benefit to giving a bit more specificity here?
- Have we standardised on camelCaseNames rather thank snake_case_names for inter-service JWT claims?

billrdunn:
> Two questions - neither of which are blockers:
> 
> * How do you feel about `previousSessionId` rather than `oldSessionId`? Is there a benefit to giving a bit more specificity here?
> * Have we standardised on camelCaseNames rather thank snake_case_names for inter-service JWT claims?

* Agree, was meaning to do that actually.
* Which line are you referring to?

mrwilson:
Disregard the last point, I've misinterpreted the code. 👍🏻

src/components/authorize/tests/authorize-service.test.ts
```diff
@@ -28,7 +30,7 @@ describe("authorize service", () => {
     process.env.API_BASE_URL = "another-base-url";
     const httpInstance = new Http();
     service = authorizeService(httpInstance);
-    getStub = sinon.stub(httpInstance.client, "get");
+    getStub = sinon.stub(httpInstance.client, "post");
```
ethanmills, line 33: This should be called `postStub` now right?

BeckaL (approved):

PR #2080, opened 2024-09-20
DFC-663: Read 'channel' claim if SUPPORT_MULTI_CHANNEL is switched-on

dbes-gds:
## What

Read 'channel' claim if SUPPORT_MULTI_CHANNEL is switched-on and assign to session variable.

- Adds a feature switch SUPPORT_MULTI_CHANNEL
- Adds a configurable default channel DEFAULT_CHANNEL, which defaults to 'web'.

This is to support rendering the UI differently per channel.  The Strategic App will run the authentication sign-in flow in a webview, which needs to be rendered differently.  The session variable can be used to determine how to render the UI.

There are two config values because the orchestration and orchestration stub changes are not yet available to supply the new claim to authentication, but the code is there to support this when it arrives.  In the meantime developers can vary the DEFAULT_CHANNEL in order to test rendering for both channels.  

## How to review

1. Code Review
1. Set DEFAULT_CHANNEL in a local frontend .env file to either 'web' or 'strategic_app' and start the application.
1. Run a journey and check log messages "Channel is set to:" to see that the channel is set appropriately, and make sure there is no impact when starting a journey.

### DIFF
--- a/ci/terraform/ecs.tf
+++ b/ci/terraform/ecs.tf
@@ -165,6 +165,14 @@ locals {
         value = var.support_check_email_fraud
       },
       {
+        name  = "SUPPORT_MULTI_CHANNEL"
+        value = var.support_multi_channel
+      },
+      {
+        name  = "DEFAULT_CHANNEL"
+        value = var.default_channel
+      },
+      {
         name  = "LANGUAGE_TOGGLE_ENABLED"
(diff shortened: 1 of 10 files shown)


BeckaL (approved):
LGTM

nickhealGDS (approved):

PR #2101, opened 2024-09-26
BAU: remove 2fa before password reset flag

BeckaL:
## What

We have now had 2fa before password reset set to on in all environments for a while and we're not likely to reverse this ever, so we can remove all unnecessary logic related to having to support this switch.

The general point of this feature flag was to ensure that users on a password reset journey had to enter mfa before being allowed to reset their passwords. Before, they were prompted to do this after resetting a password.

Best to review this commit by commit, as they build up. The general approach I've taken is incremental:

* Remove usages of the feature flag one by one from controllers, by hard-coding the places we use this feature flag to `true`;
* Remove the feature flag;
* Remove all the tf vars
* Remove the hard-coding by removing conditional logic from the state machine (we always assume the 'true' case)
* Remove hard-coding by removing conditional logic from a view (we always assume the 'true' case)
* Final sweep of the env variable from tests

## How to review

1. Remove the support 2fa before password reset variable from your env file, and run the app locally. Run through a password reset journey for 2fa and sms and see that you're prompted to enter this mfa before resetting your password. Check there's no oddities in text or errors.
1. Maybe run through a paranoid sign in journey and check everything looks normal


gtvj (approved):
I've tested the changes locally and, because I always encounter an error page at the end of journeys locally, also tested the changes on Build (having deployed the branch via the GitHub Action). Both the password reset journey and subsequent sign in worked as expected.

PR #2106, opened 2024-09-27
AUT-3763: Add support for Orch stub in addition to real Orch

ethanmills:
## What

In the build environment we want to be able to invoke the frontend using either real orchestration or the orchestration stub. To achieve this, this change allows the frontend to accept authorise requests from two different sources.

## Checklist

### DIFF
--- a/src/components/authorize/claims-config.ts
+++ b/src/components/authorize/claims-config.ts
@@ -1,4 +1,6 @@
 import {
+  getOrchStubToAuthExpectedAudience,
+  getOrchStubToAuthExpectedClientId,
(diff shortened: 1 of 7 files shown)


src/components/authorize/claims-config.ts
ethanmills, line 1: This setup makes it really hard to write tests for the validation associated with this. I've opted not to write tests for passing validation including stub values, and I don't think there were any existing tests for failed validation with incorrect values. I can restructure and add tests if needed.

dbes-gds (approved):

PR #2118, opened 2024-10-01
ATO-980: Send `authenticated` claim to backend

Ryan-Andrews99:
## What
- As [part of previous work](https://github.com/govuk-one-login/authentication-api/pull/5310/files), Orchestration now pass the `authenticated` claim to Auth as part of the `/authorize` request. We now need to pass this further to the backend so it can be stored in the new session store
- This PR pulls this claim out of the JAR and adds it to the body of the of the POST request to the start handler
- There are also some BAU commits which do some light refactoring in places:
   - Renames a test helper to use camel case
   - Updates the test helper return type to reflect more accurately the claims passed from Orchestration to Auth
   - Updates the authorize service interface for start to cut down on the number of method arguments, moving these into 
     an object.  
## How to review
- Code review commit-by-commit
- Deploy to an auth dev environment and run through a journey to test
   - Deployed to sandpit and ran through log in journey 
   - Nothing really to see, as we're not doing anything with this field yet. But as this affects the interface between Orch and 
      Auth, being successfully handed over confirms this works fine 

## Checklist

## Related PRs
- [PR which established Orch sending the authenticated claim](https://github.com/govuk-one-login/authentication-api/pull/5310/files)

### DIFF
--- a/src/components/authorize/authorize-controller.ts
+++ b/src/components/authorize/authorize-controller.ts
@@ -66,10 +66,13 @@ export function authorizeGet(
       clientSessionId,
       persistentSessionId,
       req,
-      claims.authenticated,
-      claims.reauthenticate,
-      claims.previous_session_id,
-      claims.previous_govuk_signin_journey_id
+      {
+        authenticated: claims.authenticated,
+        previous_session_id: claims.previous_session_id,
+        previous_govuk_signin_journey_id:
+          claims.previous_govuk_signin_journey_id,
+        reauthenticate: claims.reauthenticate,
+      }
     );
 
     if (!startAuthResponse.success) {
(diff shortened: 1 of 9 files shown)


BeckaL (approved):
Lovely stuff! Thanks very much for an extremely easy to understand PR with great explanations!

A v small bit of commit feedback for future - I reckon you could probably move the dial slightly away from super tiny commits - for example, your last three commits to introduce a param object are taken together small and comprehensible enough to be a single commit, and I think this probably applies to some of your earlier commits when you're e.g. changing one interface at a time to pass information through. This means that e.g. when you git blame, you've got more of the context of the change (alongside other related changes). BUT I'd much rather review something that's super easy to understand like this than a load of jumbled changes together :-D

PR #2121, opened 2024-10-02
AUT-3760: Create baseline GA4 taxonomy tests

mattvot:
## What
Work will soon be done that introduces more variation of taxonomies based on the URI, template and user session. We currently have very limited testing of existing taxonomies.

We should introduce some baseline testing of what currently exists as the upcoming work will require some refactoring.

This establishes snapshot testing of the rendered HTML on all existing integration tests that make a simulated HTTP request to the server. All pages are checked for taxonomies set and are recorded in `.snap` files.

These files are then compared in integration tests and errors thrown when they no longer match.

The implementation here adds a wrapper around the `supertest` function to run a new `expectTaxonomyMatchSnapshot` test at the end the existing expectations (this can be disabled per test).

The benefit of this means all new integration tests using the new wrapper function will automatically get taxonomy snapshot testing.

Perhaps we can remove or reduce this level of snapshot testing in time as we understand more and centralise the logic of computing taxonomies.

## How to review

1. Code Review, commit by commit and whitespace changes hidden.

### DIFF
--- a/package.json
+++ b/package.json
@@ -136,6 +136,7 @@
     "eslint-plugin-prettier": "^5.2.1",
     "husky": "^9.0.10",
     "mocha": "10.7.3",
+    "mocha-chai-jest-snapshot": "^1.1.6",
     "mock-req-res": "^1.2.0",
     "nock": "^13.5.1",
     "nodemon": "3.1.4",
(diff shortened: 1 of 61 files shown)


BeckaL (approved):
Nice one, looks good to me

PR #2144, opened 2024-10-07
AUT-3760: Refactor GA4 taxonomy setting

mattvot:
## Checklist


mattvot:
Blocking merge until performance and data analysts have approved this change

src/utils/taxonomy.ts
```diff
@@ -1,20 +1,55 @@
...
+  return taxonomyLevel2;
+}
+
+const isFeedbackTaxonomy = (req: Request): boolean => {
+  return [
+    PATH_NAMES.CONTACT_US,
+    PATH_NAMES.CONTACT_US_FROM_TRIAGE_PAGE,
+    PATH_NAMES.CONTACT_US_FURTHER_INFORMATION,
+    PATH_NAMES.CONTACT_US_QUESTIONS,
+    PATH_NAMES.CONTACT_US_SUBMIT_SUCCESS,
```
BeckaL, line 117: I'm wondering if there's any way to make this less flimsy (would be easy to miss adding this change in when we add a new contact page).

Wondering if e.g. adding a differentiation at the level of path names (grouping the contact form names into a separate object away from other path names, so that there'd be more indication when adding a new path name about where it should go), so here we'd be able to say "if in any contact us path names...". Not sure if that makes sense outside of my head!

Doesn't necessarily need fixing (if we do want to fix) in this PR, can be a follow up to reduce review complexity.

src/utils/taxonomy.ts
```diff
@@ -18,25 +24,90 @@ export type Taxonomy = {
...
+  if (isAccountInterventionTaxonomy(req))
+    taxonomyLevel2 = TaxonomyLevel2.ACCOUNT_INTERVENTION;
 
   return taxonomyLevel2;
 }
 
+const isSignInTaxonomy = (req: Request): boolean =>
+  req?.session?.user?.isSignInJourney ||
+  [
+    PATH_NAMES.ENTER_AUTHENTICATOR_APP_CODE,
```
BeckaL, line 64: I guess similar points here. But maybe this becomes a PR template checklist item for new pages - i.e. make sure the taxonomy middleware reflects the new path.

mattvot: See https://github.com/govuk-one-login/authentication-frontend/pull/2144#discussion_r1793079570

src/utils/taxonomy.ts
BeckaL: Cool, I'm happy to have it in a follow up :-)

BeckaL (dismissed):

src/utils/taxonomy.ts
gtvj: HI folks 👋🏽, thanks for seeking my input on this. I completely agree with all you've both said. One thing I do wonder about with the potential approach above where `taxonomyLevel2` is a property of the path is if there's going to be circumstances where there might not be a 1:1 relationship between a path and taxonomy values (for example in the Strategic App). I'm not entirely sure what the analytics needs will be going forward, but I'd anticipate there will be an ask for us to disambiguate between the two. This leads me to think the approach of calculating the taxonomy might give us more flexibility.

mattvot:
Having discusses this with Performance Analysts this refactor PR is OK as is.

The reauth work https://github.com/govuk-one-login/authentication-frontend/pull/2175 can be continued.

There will be follow on tickets to add taxonomy and content IDs to the following pages:
- /accessibility-statement
- /cookies
- /terms-and-conditions
- /privacy-notice

There will be content IDs added to the following pages:
- /reset-password-check-email: on error screen for code resent too many times
- /check-email-change-security-codes?type=SMS
- /check-email-change-security-codes: on error screens for 3 digits entered and incorrect 6 digits
- /enter-phone-number: on invalid number entered error
- /setup-authenticator-app: on invalid 6 digit code entered error

BeckaL (approved):

PR #2200, opened 2024-10-17
AUT-3780: Limit max node connections

alhcomer:
## What

Due to the nature of nodeJS, it is capable of handling and accepting more requests than it is able to fulfil. This can be caused through a variety of reasons, and manifests itself as increased latency in responses. We can observe this as eventLoopLag.

In an ECS context, if we allow containers to receiving too many connections, then they will just slow down and block on i/o rather than CPU - meaning that they do not report themselves as being busy against either default scaling metrics (CPUUtilized and MemUtilized).

Overload-protection is being used here to to manage to eventLoopLag. Should the eventLoop start to lag under the weight of too many connections, reject the connections with a 503 error. This prevents the container getting too busy, and aside from downstream network i/o requests made by your business logic, will stop your express apps from locking up.

The config for overload-protection in this commit was taken from [here](https://github.com/govuk-one-login/di-account-management-frontend/pull/1710). This will likely be monitored and adjusted over time. In this PR, we are deploying only to staging. I have set the logging to info so that we can see how it works when testing. This may be turned off after testing. Logging will always be turned off in production.

## How to review

1. Code Review


src/app.ts
```diff
@@ -154,6 +157,10 @@ async function createApp(): Promise<express.Application> {
 
   app.use(loggerMiddleware);
   app.use(healthcheckRouter);
+  if (getNodeEnv() === "staging") {
```
dbes-gds, line 160: let's check this is the right config variable

src/middleware/tests/overload-protection-middleware.test.ts
```diff
@@ -0,0 +1,19 @@
...
+    const result = applyOverloadProtection(isProduction);
+    expect(result).to.exist;
+  });
+
+  it("should call overloadProtection with correct options in non-production mode", () => {
+    const isProduction = false;
+    const result = applyOverloadProtection(isProduction);
+    expect(result).to.exist;
+  });
+});
```
mattvot, line 19: I think we're just asserting the middleware is returning a value, but not that it'd doing anything of value for overload protection. Happy to help figure out how to assert this with sinon

src/app.ts
dbes-gds: Please could you use getAppEnv() instead? node env is a node thing rather than our environment name.  Have a look in the ecs task definition in a deployed environment to see what these are set to.

src/middleware/overload-protection-middleware.ts
```diff
@@ -0,0 +1,15 @@
+import overloadProtection from "overload-protection";
+
+export const applyOverloadProtection = (isProduction: boolean) => {
+  return overloadProtection("express", {
```
mattvot, line 4: @alhcomer FWIW you may be able to get a return type like this:

```
export const applyOverloadProtection = (isProduction: boolean): ReturnType<typeof overloadProtection> => {
```

mattvot (approved):

PR #2280, opened 2024-11-05
BAU: Add integration test for cookies post

BeckaL:
The test itself is very basic, just testing that we can successfully make a request to the post endpoint (without testing any of the expected functionality, which is already tested and much easier to test in the unit tests). This, among other breaking changes, prevents breaking updates to the cookies library. This has recently been seen in express dependency bumps, which bumps the cookie library to 0.7.x, which contains extra regex validation of the cookie values, which we don't currently meet as we use double quotes in the value.

This should prevent re-occurences of the incident that was caused by this cookie library bump.

## How to review

For example:

1. Code Review
1. While on this branch, bump a dependency locally such that cookies is updated to 0.7.x (for example, by bumping express to 0.4.21 as done in [this reverted PR.
](https://github.com/govuk-one-login/authentication-frontend/pull/2265) Verify that the new test fails, preventing us from accidentally bumping this dependency without being alerted that we need to implement a fix

### DIFF
--- a/src/components/common/cookies/tests/cookies-controller-integration.test.ts
+++ b/src/components/common/cookies/tests/cookies-controller-integration.test.ts
@@ -10,54 +10,88 @@ describe("Integration:: cookies controller", () => {
   const dynamicGACookieName = "_gat_UA-[number]";
   const analyticsCookieNamesListedInCookieNotice: string[] = [];
 
-  before(async () => {
-    decache("../../../../app");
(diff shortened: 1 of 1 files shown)


dbes-gds (approved):

gtvj (approved):
Thanks @BeckaL - I've tested it as you suggested (by bumping Express) and the test does fail as expected.

PR #2309, opened 2024-11-12
AUT-3834: add basic ipv callback endpoint

BeckaL:
## What

Adds a very basic controller and route, in the case that the mfa reset with IPV feature flag is on, that responds to a callback from the authorize endpoint in IPV, and for now just responds with a basic success status. This is to enable step by step implementation, so that we can see the callback successfully working.

## How to review

1. Code Review

### DIFF
--- a/src/components/ipv-callback/ipv-callback-routes.ts
+++ b/src/components/ipv-callback/ipv-callback-routes.ts
@@ -1,0 +1,10 @@
+import * as express from "express";
+import { asyncHandler } from "../../utils/async";
+import { PATH_NAMES } from "../../app.constants";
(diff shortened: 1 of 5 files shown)


src/components/ipv-callback/ipv-callback-routes.ts
```diff
@@ -0,0 +1,10 @@
+import * as express from "express";
```
BeckaL, line 1: have a look at this

BeckaL: actually ignore this is the pattern we use

alhcomer (approved):

PR #2324, opened 2024-11-18
Aut 3824: Implement authentication callback

BeckaL:
## What

Adds to the existing dummy callback endpoint, to call the reverification result lambda. This is the next step in implementing the mfa reset with identity journey, and covers off step 10/11 in the flow diagram linked in the epic.

## How to review

For example:

1. Code Review
1. Deploy to dev alongside [this api branch](https://github.com/govuk-one-login/authentication-api/tree/AUT-3824/temporarily-hardcode-success-in-reverification-result). 
1. Run through an mfa reset journey with the network tab open on dev tools. See that on callback from the stub to the frontend, we now make a further call to the backend's reverification endpoint, and that this is successful and successfully displays the "reverification result success" message added in this PR. Verify the reverification result lambda was called in the aws console if non obvious from dev tools.

### DIFF
--- a/src/components/ipv-callback/ipv-callback-controller.ts
+++ b/src/components/ipv-callback/ipv-callback-controller.ts
@@ -1,8 +1,28 @@
 import { Request, Response } from "express";
 import { ExpressRouteFunc } from "../../types";
+import { ReverificationResultInterface } from "./types";
+import { logger } from "../../utils/logger";
+import { reverificationResultService } from "./reverification-result-service";
 
-export function ipvCallbackGet(): ExpressRouteFunc {
+export function ipvCallbackGet(
+  service: ReverificationResultInterface = reverificationResultService()
+): ExpressRouteFunc {
   return async function (req: Request, res: Response) {
-    res.status(200).send("Got to ipv callback");
+    const { email } = req.session.user;
+    const { sessionId, clientSessionId, persistentSessionId } = res.locals;
+
+    const result = await service.getReverificationResult(
+      sessionId,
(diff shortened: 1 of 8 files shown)


src/components/ipv-callback/types.ts
```diff
@@ -0,0 +1,12 @@
...
+import { Request } from "express";
+
+export interface ReverificationResultInterface {
+  getReverificationResult: (
+    sessionId: string,
+    clientSessionId: string,
+    persistentSessionId: string,
+    req: Request,
+    email: string
+  ) => Promise<ApiResponseResult<DefaultApiResponse>>; //TODO add type
```
BeckaL: This TODO covered off in later commit

src/components/ipv-callback/ipv-callback-controller.ts
```diff
@@ -1,8 +1,42 @@
...
+    const result = await service.getReverificationResult(
+      sessionId,
+      clientSessionId,
+      persistentSessionId,
+      req,
+      email,
+      code
+    );
+
+    logger.info(
```
andrew-moores, line 32: Do we need to log out when this is successful or just when a User fails to be re-verified?

andrew-moores (approved):

src/components/ipv-callback/ipv-callback-controller.ts
BeckaL: I think for now useful to log in both cases because this is the only feedback in logs - but when we implement more of the flow we can probably just log in the case of non success

PR #2447, opened 2025-01-08
BAU: Cloudfront WAF improvements

whi-tw:
## What

Details in the individual commits.

TL;DR: rate limits now apply to endpoints individually, config format has changed

## How to review

- Code Review

1. Deploy to an environment without making any changes to the tfvars: observe that the configuration is translated to the new format
2. Update the values of `{aps_session,ip}_endpoint_rate_limiting_configuration` tfvars, observe that your new values are used instead
3. I guess you could do some brute-forcing, to check whether the rules actually work. If the syntax is right, they will work, though.

### DIFF
--- a/ci/terraform/variables.tf
+++ b/ci/terraform/variables.tf
@@ -307,25 +307,101 @@ variable "alb_idle_timeout" {
 }
 
 variable "rate_limited_endpoints" {
-  description = "List of endpoints that should be rate limited by session and IP"
+  description = "**DEPRECATED** List of endpoints that should be rate limited by session and IP"
   type        = list(string)
-  default     = ["/dummy-77349847-9ce0-499c-b378-6d9b49a24d6a", "/dummy-168a5a48-62f9-44b1-b9ea-d4c8a74b8498"] # default to two uuids, which are not real endpoints. This prevents us from having an empty or statement
-  validation {
-    condition     = length(var.rate_limited_endpoints) >= 2
-    error_message = "rate_limited_endpoints must contain at least two endpoints. If you only have one endpoint, add a non-existent dummy, eg. `/dummy-77349847-9ce0-499c-b378-6d9b49a24d6a`"
-  }
+  default     = []
 }
 
 variable "rate_limited_endpoints_rate_limit_period" {
-  description = "Period in seconds for rate limiting for rate limited endpoints"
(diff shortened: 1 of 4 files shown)


ci/terraform/variables.tf
```diff
@@ -307,27 +307,103 @@ variable "alb_idle_timeout" {
...
+  }
+  validation {
+    condition     = length(var.ip_endpoint_rate_limiting_configuration) == 0 || alltrue([for config in var.ip_endpoint_rate_limiting_configuration : config.limit >= 10])
+    error_message = "limit must be >= 10."
+  }
+}
+
+locals {
+  // for transition to new-style configuration. remove once all environments have had variables updated
+  ip_endpoint_rate_limiting_configuration = length(var.aps_session_endpoint_rate_limiting_configuration) > 0 ? var.ip_endpoint_rate_limiting_configuration : [
```
mattvot, line 358: Should this be `length(var.ip_endpoint_rate_limiting_configuration) > 0`, not `aps`?

ci/terraform/waf.tf
```diff
@@ -83,114 +83,205 @@ resource "aws_wafv2_web_acl" "frontend_cloudfront_waf_web_acl" {
...
+                        priority = 1
+                        type     = "LOWERCASE"
+                      }
+                    }
                   }
                 }
               }
             }
+            dynamic "byte_match_statement" {
+              // If there's only 1 endpoint, just create the single byte_match_statement
```
mattvot, line 139: I presume this duplication is because `or_statement` doesn't support just 1 `byte_match_statement`

mattvot (changes requested):

ci/terraform/variables.tf
whi-tw: Indeed it should!

ci/terraform/waf.tf
whi-tw: Correct. Previously we had to add a second 'fake' endpoint to protect for the or to work, but now it'll gracefully handle a single entry without that hack

ci/terraform/variables.tf
whi-tw: resolved!

mattvot (approved):

PR #2449, opened 2025-01-08
AUT-3842: Conditionally changing messaging on the Session Expiry Error page

ben-fernandes-sw:
## What

Different text is shown on the session expiry page depending on whether it's rendering within the app or not.

The CTA shown normally relies on the context of being on the web, so was previously confusing when displayed in the app. This changes the text and removes the CTA button.

### Existing content

| English | Welsh |
| --- | --- |
| existing-english | existing-welsh |

### New (app-specific) content

| English | Welsh |
| --- | --- |
| new-english | new-welsh |

### New `channel` cookie

| English | Welsh |
| --- | --- |
| Screenshot 2025-01-20 at 11 08 22 | Screenshot 2025-01-20 at 11 10 06 |

## How to review

1. Code Review
1. Test locally
  - You'll need to trick the frontend into thinking that you're in an expired app session.
    - To make it think you're in an app session, modify the `channel-middleware.ts` file to set `res.locals.strategicAppChannel = true` and `res.locals.webChannel = false`
    - To make it think you're in an expired session, modify the `session-middleware.ts` file to use a nonsense `sessionId`. This will force the session expiry page.
  - Test that the shortened text is displayed as above, including after a refresh, and that the language toggle works correctly.

## Checklist

### DIFF
--- a/src/middleware/channel-middleware.ts
+++ b/src/middleware/channel-middleware.ts
@@ -1,13 +1,23 @@
 import { NextFunction, Request, Response } from "express";
-import { CHANNEL } from "../app.constants";
+import { CHANNEL, COOKIES_CHANNEL } from "../app.constants";
 
 export function channelMiddleware(
   req: Request,
   res: Response,
   next: NextFunction
 ): void {
-  res.locals.strategicAppChannel =
-    req.session?.user?.channel === CHANNEL.STRATEGIC_APP;
-  res.locals.webChannel = req.session?.user?.channel === CHANNEL.WEB;
+  if (req.session?.user?.channel) {
+    setChannelFlags(res, req.session.user.channel);
+  } else if (req.cookies[COOKIES_CHANNEL]) {
+    setChannelFlags(res, req.cookies[COOKIES_CHANNEL]);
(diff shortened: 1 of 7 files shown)


BeckaL (dismissed):
Lovely stuff, LGTM 🥳

src/middleware/channel-middleware.ts
```diff
@@ -1,13 +1,23 @@
...
 ): void {
-  res.locals.strategicAppChannel =
-    req.session?.user?.channel === CHANNEL.STRATEGIC_APP;
-  res.locals.webChannel = req.session?.user?.channel === CHANNEL.WEB;
+  if (req.session?.user?.channel) {
+    setChannelFlags(res, req.session.user.channel);
+  } else if (req.cookies[COOKIES_CHANNEL]) {
+    setChannelFlags(res, req.cookies[COOKIES_CHANNEL]);
+  } else {
+    setChannelFlags(res, "web");
```
dbes-gds, line 14: mind using a constant here?

dbes-gds (approved):

PR #2460, opened 2025-01-10
AUT-3920: Redirect back to enter mfa pages

alhcomer:
## What

- If a user selects "Try entering a security code again with the method you already have set up" in the cannot-change-how-get-security-codes page they will be redirected to the OTP challenge screen again. They are not sent another MFA code.
- Enter mfa attempt counts persist for both MFA types on the back end.
- Mfamethod type has been stored on a user's session from the enter password stage where it is retrieved from the db in order to know which OTP challenge screen they will be redirected to.

## How to review

1. Code Review
2. Tail cloud watch logs for chosen MFA method handler (VerifyMfaCodeHandler for auth app, VerifyCodeHandler for SMS)
3. Run against dev and run through a sign in journey up to MFA challenge, enter an invalid code, then select 'Problems with the code?' -> 'change how you get security codes'
4. Select error response in IPV stub and continue
5. Enter invalid code again
6. Open new stub and go through sign in journey up to MFA, entering an incorrect code again
7. See that the count has been incorrect code count has been incremented from 0 to 3 in cloudwatch 

## Checklist

### DIFF
--- a/src/components/common/state-machine/state-machine.ts
+++ b/src/components/common/state-machine/state-machine.ts
@@ -761,10 +761,20 @@ const authStateMachine = createMachine(
         },
       },
       [PATH_NAMES.CANNOT_CHANGE_SECURITY_CODES]: {
-        type: "final",
+        on: {
+          [USER_JOURNEY_EVENTS.VERIFY_MFA]: [PATH_NAMES.ENTER_MFA],
+          [USER_JOURNEY_EVENTS.VERIFY_AUTH_APP_CODE]: [
+            PATH_NAMES.ENTER_AUTHENTICATOR_APP_CODE,
+          ],
+        },
       },
       [PATH_NAMES.CANNOT_CHANGE_SECURITY_CODES_IDENTITY_FAIL]: {
-        type: "final",
+        on: {
+          [USER_JOURNEY_EVENTS.VERIFY_MFA]: [PATH_NAMES.ENTER_MFA],
+          [USER_JOURNEY_EVENTS.VERIFY_AUTH_APP_CODE]: [
+            PATH_NAMES.ENTER_AUTHENTICATOR_APP_CODE,
+          ],
+        },
       },
     },
   },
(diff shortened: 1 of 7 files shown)


src/components/ipv-callback/tests/ipv-callback-integration.test.ts
```diff
@@ -180,5 +161,98 @@ describe("Integration:: ipv callback", () => {
...
+        email: "test@test.com",
+        phoneNumber: "7867",
+        journey: {
+          nextPath: PATH_NAMES.CANNOT_CHANGE_SECURITY_CODES,
+        },
+        mfaMethodType: mfaMethodType,
+      };
+
+      next();
+    });
```
mattvot: I think this could move to `getCannotChangeSecurityCodesAndReturnTokenAndCookies`, as this is done just to facilitate getting the right token and cookies set right?

Maybe difficult though, and how we stub it would probably need to change. So no requirement to modify this now.

src/components/common/state-machine/state-machine.ts
```diff
@@ -761,7 +761,12 @@ const authStateMachine = createMachine(
         },
       },
       [PATH_NAMES.CANNOT_CHANGE_SECURITY_CODES]: {
-        type: "final",
+        on: {
+          [USER_JOURNEY_EVENTS.VERIFY_MFA]: [PATH_NAMES.ENTER_MFA],
+          [USER_JOURNEY_EVENTS.VERIFY_AUTH_APP_CODE]: [
+            PATH_NAMES.ENTER_AUTHENTICATOR_APP_CODE,
+          ],
+        },
```
mattvot, line 769: This should also be supported on the path below too, `CANNOT_CHANGE_SECURITY_CODES_IDENTITY_FAIL`

mattvot (changes requested):

src/components/ipv-callback/tests/ipv-callback-integration.test.ts
alhcomer: I had crack at this but found it fiddly as it only works when the app is created after the stubbing. This is the area (along with decaching) that I don't fully understand

mattvot (approved):

PR #2491, opened 2025-01-17
BAU: Fix use of .env file for the dev-x.js stubs

ben-fernandes-sw:
## What

This contains two changes:

### Fix .env file not being read into dev-x local stubs

The dev-x.js files use dotenv to load environment variables. Therefore, they expect a .env file to exist in the /app directory in the Docker container. We weren't copying in the .env file, so the scripts were always falling back onto default values.

### dev-x stubs use SESSION_EXPIRY as defined in .env file

There were two issues previously:
1) When setting cookies in the stubs, we set the 'maxAge' as a datetime 1 hour in the future. This didn't have the intended effect though, as maxAge expects an integer of milliseconds, not a date as we were providing. 'expire' should be used to set expiry at a specific datetime.
2) This is a moot point now though. The session expiry was hardcoded to an hour in the future. Now, it uses the SESSION_EXPIRY param as defined in the .env file.

The 'lng' cookie is still set to a year in the future, but that's because a library we use overwrites it.

## How to review

1. Code Review

### DIFF
--- a/dev-app.js
+++ b/dev-app.js
@@ -12,6 +12,7 @@ const logger = pino({ level: process.env.LOG_LEVEL || "info" });
 const app = express();
 const port = process.env.PORT || 3001;
 const frontend_port = process.env.FRONTEND_PORT || 3000;
+const sessionExpiry = Number(process.env.SESSION_EXPIRY || 3600000);
 
 function createAuthorizeRequest() {
   const vtr = process.env.VTR ? `vtr=${encodeURI(process.env.VTR)}&` : "";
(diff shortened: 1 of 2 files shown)


dev-app.js
```diff
@@ -62,7 +63,7 @@ app.get("/", (req, res) => {
       if (lngCookie) {
         lngCookieValue = getCookieValue(lngCookie.split(";"), "lng");
         res.cookie("lng", lngCookieValue, {
-          maxAge: new Date(new Date().getTime() + 60 * 60000),
```
dbes-gds, line 65: maxAge rather than expires, currently falling over for me as expires takes a date

dbes-gds (approved):

PR #2521, opened 2025-01-27
AUT-3957: Upgrade analytics to support 5 taxonomy levels

mattvot:
## What

`@govuk-one-login/frontend-analytics` is updated to 3.1.0 and support added for the 3 new taxonomy levels

`taxonomy.ts` is updated to support empty strings as default for each new level, with future work expanding on these possible values.

Note that it is intended that there are a lot of empty strings sent to the GA4 library, which matches the implementation at https://github.com/govuk-one-login/govuk-one-login-frontend/blob/60326ab8f4d5c77e7f7ede8c0030431ab4b614f3/packages/frontend-analytics/src/components/ga4-opl/template.njk#L12. The GA4 library itself strips empty string values from the network requests to Google Analytics.

## How to review

1. Code Review

## Related PRs

Builds off the back of https://github.com/govuk-one-login/authentication-frontend/pull/2520

### DIFF
--- a/package.json
+++ b/package.json
@@ -67,7 +67,7 @@
     "@aws-sdk/client-ssm": "^3.366.0",
     "@eslint/eslintrc": "^3.1.0",
     "@eslint/js": "^9.14.0",
-    "@govuk-one-login/frontend-analytics": "3.0.1",
+    "@govuk-one-login/frontend-analytics": "3.1.0",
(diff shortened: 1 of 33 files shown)


BeckaL (approved):
LGTM, helpful refactor of that test file too, clearer for me now

BeckaL (approved):

PR #2524, opened 2025-01-28
AUT-3993: Tell backend not to block user for mfa reset when mfa reset via ipv is on

BeckaL:
## What

This makes use of the new optional flag ("allowMfaResetAfterPasswordReset") on the request to the reset password handler when the mfa reset via ipv journey is on.

This will mean that a user who has successfully reset their password does not have any account modifiers put on them, and can immediately reset their mfa on a subsequent journey after resetting their password (previously they would have had to complete one successful subsequent login journey before resetting mfa.

This behaviour was somewhat legacy anyway since we put the MFA challenge as part of password reset, but will be especially now that they have to go through a higher security journey as part of resetting their MFA.

We've made a decision here not to introduce a separate feature flag on the backend that reflects whether the new support mfa journey is on or not. Instead the behaviour is controlled via the frontend feature flags, meaning this new behaviour will all be switched on at once.

A consequence of this is that if, once we're happy with the behaviour in production, we want to remove the feature flag and have the 'on' behaviour as default, we'll need to first default the behaviour in the backend (or always set this value to true, and then default the backend behaviour). This should be obvious when changing the code, but I've added in a rare comment to be on the safe side.

Also contains some test tidies.

## How to review

* Code review commit by commit - might be best to turn off whitespace comparison as I've added some nesting to tests which looks nasty without this
* QA will cover off testing but if you wanted to test you would deploy this and the frontend change to an env with the new journey flagged on or off, and see that either you could reset your MFA in the journey following a password reset (when the new journey is switched on), or you could not if the new journey is switched off.

## Related PRs

Requires [backend changes](https://github.com/govuk-one-login/authentication-api/pull/5798) to be merged first

### DIFF
--- a/src/components/reset-password/reset-password-controller.ts
+++ b/src/components/reset-password/reset-password-controller.ts
@@ -11,6 +11,7 @@ import { USER_JOURNEY_EVENTS } from "../common/state-machine/state-machine";
 import { BadRequestError } from "../../utils/error";
 import { EnterPasswordServiceInterface } from "../enter-password/types";
 import { enterPasswordService } from "../enter-password/enter-password-service";
+import { supportMfaResetWithIpv } from "../../config";
 
 const resetPasswordTemplate = "reset-password/index.njk";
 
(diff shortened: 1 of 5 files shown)


src/components/reset-password/reset-password-controller.ts
```diff
@@ -48,6 +49,9 @@ export function resetPasswordPost(
       clientSessionId,
       persistentSessionId,
       withinForcedPasswordResetJourney,
+      // WARNING! Before removing this feature flag, you first need to ensure that this behaviour is the default on the backend
+      // Purely removing this feature flag will mean that we start writing to account modifiers again
+      supportMfaResetWithIpv(),
```
andrew-moores, line 54: Could this be extracted and a conditional added for clarity:

const allowMfaResetAfterPasswordReset = supportMfaResetWithIpv(): true ? false

and then use the variable in the call to updatePassword.

BeckaL: Sounds good, fixed up in the commit now

andrew-moores (approved):

andrew-moores (approved):

PR #2525, opened 2025-01-28
AUT-3964: Disable, remove and clean up Universal Analytics code

ben-fernandes-sw:
### The problem
✅ GA4 events were unaffected
✅ UA events stopped as expected, but nothing was listening for them anyway
⚠️ RUM events also stopped - but why??

### The reason
In the `authentication-frontend` repo we have an `application.js` file that initialises UA and RUM. We set the `analyticsUa.init` function to call our own UA init (via the `cookies.initAnalytics` function), but we also hijack it to initialise RUM. `analyticsUa.init` is called by the analytics library, but only if UA is enabled. Therefore, by disabling UA in the library, we disabled calling the `analyticsUa.init` function, so disabled RUM init.

### The fix
Split apart the RUM and UA code so that they don't depend on each other, so they init on page load, and also whenever the cookie consent state changes (the analytics library has an event emitted for this). Determine the analytics consent state by its cookie.

There's a fair bit of UA code dotted around the repo that's confusing things (Terraform / CloudFormation / `application.js` / `cookies.js`) - this PR removes and cleans up that code too, now that we don't need it any more.

## How to review

1. Code Review
1. Deploy and test that
    1. Analytics events are _not_ emitted on first load of the page
    2. Analytics events are emitted after the "Accept" button is clicked
    3. Analytics events continue to be emitted on page refresh
    4. If analytics are disabled on the `/cookies` page, the page should refresh and events should stop being emitted
    5. If the user clicks "Reject" analytics events were never emitted

### DIFF
--- a/ci/tasks/deploy-frontend.yml
+++ b/ci/tasks/deploy-frontend.yml
@@ -11,12 +11,11 @@ params:
   STATE_BUCKET: digital-identity-dev-tfstate
   STATE_LOCKING_TABLE: digital-identity-dev-tfstate-locking
   DEPLOY_ENVIRONMENT: build
-  GTM_ID: ((build-gtm-id))
   SESSION_EXPIRY: ((build-session-expiry))
   BASIC_AUTH_USERNAME: ((no-basic-auth-username))
   BASIC_AUTH_PASSWORD: ((no-basic-auth-password))
   INCOMING_TRAFFIC_CIDR_BLOCKS: '["0.0.0.0/0"]'
-  BASIC_AUTH_BYPASS_CIDR_BLOCKS: '[]'
+  BASIC_AUTH_BYPASS_CIDR_BLOCKS: "[]"
 inputs:
   - name: frontend-src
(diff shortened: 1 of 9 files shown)


src/components/common/layout/base.njk
```diff
@@ -174,11 +174,10 @@
                     .DI
                     .appInit({
                         ga4ContainerId: "{{ga4ContainerId}}",
-                        uaContainerId: "{{uaContainerId}}"
+                        uaContainerId: "{{uaContainerId}}" // We don't use UA anymore, but it's a required param
```
andrew-moores, line 177: Can we pass a dummy string, `not used`, if we can't change the signature of appInit?

andrew-moores (commented):
No tests for the new functions in `application.js`?

andrew-moores (dismissed):
Tested the GA part.  Code reviewed the DynaTrace part as we don't have RUM available for testing.

andrew-moores (approved):

PR #2557, opened 2025-02-06
AUT-3846: Remove basic auth

whi-tw:
## What

Remove basic auth - no longer a requirement.

Update the banner, to indicate that these are not the droids you're looking for.

## How to review

- Ensure no goofs in removal
- Ensure it's obvious that non-production environments are non-production. TBC.


whi-tw:
image

dbes-gds:
Anything need doing on the secure pipeline/CF side?

.github/workflows/pre-commit.yml
```diff
@@ -37,7 +37,7 @@ jobs:
         run: yarn install
 
       - name: Run pre-commit
-        uses: pre-commit/action@2c7b3805fd2a0fd8c1884dcaebf91fc102a13ecd # v3.0.1
+        uses: whi-tw/pre-commit-action@9a134f4f53ddd4071215facfc20ea754f9cb337b # v3.1.1
```
andrew-moores, line 40: Should this be owned by OneLogin rather than you?

whi-tw: the previous one wasn't, so I don't really see why! It's sha-pinned, so I can't rug y'all if I leave.

whi-tw: the previous one wasn't, so I don't really see why! It's sha-pinned, so I can't rug y'all if I leave.  idk.

whi-tw:
Have tested it deploys sensibly to both TF and CF environments:

- TF: https://signin.authdev1.sandpit.account.gov.uk/sign-in-or-create
- CF: https://signin.authdev1.dev.account.gov.uk/sign-in-or-create

andrew-moores (approved):

PR #2632, opened 2025-03-07
ATO-1523: Remove doc-app code linked to auth backend

cearl1:
## What

We are looking to remove some of the doc-app related code from the auth backend. This means we will no longer be sending a `docCheckingAppUser` boolean back to the frontend. I'm not sure what would happen if we stopped sending this field to the frontend, so I've raised this PR to remove references to this field in the response from the backend.

Since the split, the front-end should only be handling auth journeys, so this flag should already always be false.


src/components/authorize/authorize-controller.ts
```diff
@@ -129,7 +129,6 @@ export function authorizeGet(
         isIdentityRequired: req.session.user.isIdentityRequired,
         isAuthenticated: req.session.user.isAuthenticated,
         prompt: req.session.client.prompt,
-        skipAuthentication: req.session.user.docCheckingAppUser,
```
cearl1, line 132: This is defaulted to false in the state machine

cearl1: To make this clearer, i've reverted this change and set the skipAuthentication to `false`

Ryan-Andrews99 (approved):

PR #2660, opened 2025-03-17
AUT-4114: Remove /prove-identity endpoint

ben-fernandes-sw:
## What

This endpoint is no longer used, so it can be completely removed. This PR removes it as a valid path name, +from the state machine, +as a controller route.

## How to review

1. Code Review

### DIFF
--- a/src/components/prove-identity-callback/tests/prove-identity-callback-service.test.ts
+++ b/src/components/prove-identity-callback/tests/prove-identity-callback-service.test.ts
@@ -49,7 +49,7 @@ describe("prove identity callback service", () => {
       postStub.resolves(axiosResponse);
       const { email, sessionId, clientSessionId, diPersistentSessionId } =
         commonVariables;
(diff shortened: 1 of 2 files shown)


src/components/prove-identity-callback/tests/prove-identity-callback-service.test.ts
```diff
@@ -49,7 +49,7 @@ describe("prove identity callback service", () => {
       postStub.resolves(axiosResponse);
       const { email, sessionId, clientSessionId, diPersistentSessionId } =
         commonVariables;
-      const req = createMockRequest(PATH_NAMES.PROVE_IDENTITY, {
+      const req = createMockRequest("/testPath", {
```
ethanmills, line 52: This is still `/prove-identity` because that's the path on the API

mattvot (dismissed):

mattvot (approved):

PR #2697, opened 2025-03-31
AUT-4209: Cache static assets

ethanmills:
## What

Add some more permissive cache control headers to the static assets. 

## How to review
Ensure application pages remain uncached

## Checklist
No relevant items

### DIFF
--- a/src/components/common/layout/tests/base-integration.test.ts
+++ b/src/components/common/layout/tests/base-integration.test.ts
@@ -173,4 +173,14 @@ describe("Integration:: base page ", () => {
       expect($("a.govuk-header__link").length).to.equal(1);
     });
   });
+
+  describe("Cache-Control", () => {
+    it("should not be cached", async () => {
+      await setupApp("");
+      const response = await request(app).get(PATH_NAMES.SIGN_IN_OR_CREATE);
+      expect(response.header["cache-control"]).to.equal(
+        "no-cache, private, no-store, must-revalidate, max-stale=0, post-check=0, pre-check=0"
+      );
+    });
+  });
 });
--- a/src/app.ts
+++ b/src/app.ts
@@ -1,6 +1,7 @@
 import express, { Application } from "express";
 import cookieParser from "cookie-parser";
 import csurf from "csurf";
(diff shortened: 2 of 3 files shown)


ethanmills:
I've tested this in dev

src/components/common/layout/tests/base-integration.test.ts
```diff
@@ -173,4 +173,14 @@ describe("Integration:: base page ", () => {
       expect($("a.govuk-header__link").length).to.equal(1);
     });
   });
+
+  describe("Cache-Control", () => {
+    it("should not be cached", async () => {
+      await setupApp("");
+      const response = await request(app).get(PATH_NAMES.SIGN_IN_OR_CREATE);
+      expect(response.header["cache-control"]).to.equal(
+        "no-cache, private, no-store, must-revalidate, max-stale=0, post-check=0, pre-check=0"
```
cjfryer, line 182: My reading of the spec is that `no-store` achieves the intent of this, i.e. "caches of any kind (private or shared) should not store this response."  I have a feeling the rest are superfluous (but you may be working around broken browser behaviour I don't know about.)

ethanmills: It seems to have been this way since basically the dawn of time, not sure if there's a good reason or not https://github.com/govuk-one-login/authentication-frontend/commit/de5dce16a351ac6b458840da628d8dc40b757043#diff-2442f93598ecc4ba3036deaaaf60315162f86ca95b2c42ea10671a2b973c693fR10

dbes-gds (approved):

src/components/common/layout/tests/base-integration.test.ts
cjfryer: Thanks. Definitely no argument for changing it here, then.  I see these "kitchen sink" cache-control headers in the wild all the time, so they probably don't hurt.  We should take a look when we finally get round to [defining the programme approach](https://github.com/govuk-one-login/architecture/blob/main/rfc/0072-assets-via-cloudfront.md#step-3-caching-of-static-assets)

PR #2702, opened 2025-04-02
ATO-1435: Pass new claims to auth backend.

cearl1:
Reopening PR after fixing issues with orch stub

## What

Now we are passing new fields from orch to the auth frontend, we would like to pass these fields to the auth backend, to avoid having to use the client session authRequestParams field. 

Also added tests to ensure we only pass optional parameters to the backend when they are sent to frontend by orch
## How to review

1. Code Review

Deployed to sandpit + authdev1 with all other PRs, verified that the claims that are passed from orch to auth backend were in sync with the client session auth request

## Checklist

- [x] Performance analyst has been notified of the change. n/a
- [x] A UCD review has been performed. n/a
- [x] Any necessary changes to the [acceptance tests](https://github.com/govuk-one-login/authentication-acceptance-tests) have been made. n/a 
- [x] Documentation has been updated to reflect these changes. n/a

### DIFF
--- a/src/components/authorize/authorize-service.ts
+++ b/src/components/authorize/authorize-service.ts
@@ -72,5 +72,14 @@ function createStartBody(startRequestParameters: StartRequestParameters) {
   )
     body["previous-govuk-signin-journey-id"] =
       startRequestParameters.previous_govuk_signin_journey_id;
+  if (startRequestParameters.cookie_consent !== undefined)
+    body["cookie_consent"] = startRequestParameters.cookie_consent;
+  if (startRequestParameters._ga !== undefined)
+    body["_ga"] = startRequestParameters._ga;
+  body["vtr"] = startRequestParameters.vtr;
+  body["client_id"] = startRequestParameters.client_id;
+  body["scope"] = startRequestParameters.scope;
+  body["redirect_uri"] = startRequestParameters.redirect_uri;
+  body["state"] = startRequestParameters.state;
   return body;
 }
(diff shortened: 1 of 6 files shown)


Louisasa (approved):

src/components/authorize/authorize-service.ts
```diff
@@ -72,5 +72,14 @@ function createStartBody(startRequestParameters: StartRequestParameters) {
...
       startRequestParameters.previous_govuk_signin_journey_id;
+  if (startRequestParameters.cookie_consent !== undefined)
+    body["cookie_consent"] = startRequestParameters.cookie_consent;
+  if (startRequestParameters._ga !== undefined)
+    body["_ga"] = startRequestParameters._ga;
+  body["vtr"] = startRequestParameters.vtr;
+  body["client_id"] = startRequestParameters.client_id;
+  body["scope"] = startRequestParameters.scope;
+  body["redirect_uri"] = startRequestParameters.redirect_uri;
+  body["state"] = startRequestParameters.state;
```
Ryan-Andrews99, line 83: Should we be passing the rp client id here?

cearl1: Yeah its a bit hard to see but we transform it from `rp_client_id` to `client_id` when we make the request to the backend in the authorize-controller: 
```
client_id: claims.rp_client_id,
```

Ryan-Andrews99: Ooooh I see what you mean

PR #2714, opened 2025-04-07
ATO-1435: Pass new claims to auth backend.

cearl1:
## What

We have defined some new claims which will replace the need for auth client session

This PR passes these new claims from orch to the auth backend.

## How to review

Deployed to sandpit with the backend changes, and completed an auth journey.

## Checklist

## Related PRs

Would need to merge the backend changes first: https://github.com/govuk-one-login/authentication-api/pull/6297


src/components/authorize/authorize-controller.ts
```diff
@@ -74,9 +73,16 @@ export function authorizeGet(
...
           claims.previous_govuk_signin_journey_id,
         reauthenticate: claims.reauthenticate,
+        cookie_consent: claims.cookie_consent,
+        _ga: claims._ga,
+        scope: claims.scope,
+        client_id: claims.rp_client_id,
+        redirect_uri: claims.rp_redirect_uri,
+        state: claims.rp_state,
+        requested_credential_strength: claims.requested_credential_strength,
+        requested_level_of_confidence: claims.requested_level_of_confidence,
```
Ryan-Andrews99, line 83: I think we could just pass the claims in here instead of doing all this pulling out? I think I wrote this initially so kind of my fault 😄

Ryan-Andrews99 (commented):
Think we should do a bit more manual testing with this one in sandpit, just checking things like:
- No VTR param
- VTR with requested_level_of_confidence
Just to make sure things are in sync

src/components/authorize/authorize-controller.ts
cearl1: We are renaming some of the fields though (`rp_client_id` becomes `client_id`, same with `redirect_uri` and `state`). I could use the spread operator and set the renamed fields explicitly - this would mean that we are sending some fields to the backend twice though.

Ryan-Andrews99: Ahh I see what you mean, maybe we can leave it

cearl1: I've realized we can move the renaming of these fields into the authorize-service call. I've added a separate commit specifically for this change, just so you can see where the renaming has moved to a bit clearer

Ryan-Andrews99 (approved):

PR #2738, opened 2025-04-16
AUT-4254: Forbid navigation to root

dbes-gds:
## What

Forbid navigation to root

Before the Auth Orch split the front-door authorize endpoint lived on the backend.  This redirected to the frontend root (e.g. signin.account.gov.uk) in order to start the frontend journey, so root was the landing page and had corresponding entries in the state machine.

Since the Auth authorize endpoint moved to frontend root became redundant as the starting point of the frontend journey is now /authorize itself.

This PR removes legacy code supporting root.  If a user tries to navigate to root they will now be shown the 'Sorry, you cannot access GOV.UK One Login from this page' error page with an http 403 status.

> Screenshot 2025-04-16 at 15 13 47

## How to review

1. Code Review
1. Deploy to an environment for testing.
1. Start a journey and go through a few steps
1. Navigate manually to root to see that the correct error page is displayed.


CarlyG55:
Happy with this approach, will let someone who knows the finer detail of the code do a full review 🙂

BeckaL (approved):

PR #2791, opened 2025-05-19
AUT-4297: Add mobile channel

alhcomer:
## What

Accept 'generic_app' as a valid channel on the FE and render the same pages and content as the strategic app when res.locals.genericApp is true. This is done by referencing the new res.locals.isApp parameter in templates, which is true when either of the aforementioned res.locals params are true.

## How to review

1. Code Review

## Checklist

### DIFF
--- a/src/components/account-created/account-created-controller.ts
+++ b/src/components/account-created/account-created-controller.ts
@@ -7,7 +7,6 @@ export function accountCreatedGet(req: Request, res: Response): void {
   res.render("account-created/index.njk", {
     serviceType,
     name,
-    strategicAppChannel: res.locals.strategicAppChannel,
   });
 }
 
--- a/src/components/common/layout/base.njk
+++ b/src/components/common/layout/base.njk
@@ -6,7 +6,7 @@
 
 {% from "../header/macro.njk" import strategicAppHeader %}
 
-{% if strategicAppChannel == true %}
+{% if isApp %}
   {% set htmlClasses = 'govuk-template__mobile' %}
 {% endif %}
 
@@ -53,7 +53,7 @@
 {% set phaseBannerClasses  = "test-banner" if showTestBanner %}
 
 {% block header %}
-    {% if strategicAppChannel %}
+    {% if isApp %}
     {{ strategicAppHeader({
         useTudorCrown: true,
         classes: phaseBannerClasses
@@ -90,7 +90,7 @@
 
 {% block main %}
     <div class="govuk-width-container {{ containerClasses }}">
-        {% if strategicAppChannel === true %}
+        {% if isApp %}
         {% else %}
         {{ govukPhaseBanner({
             tag: {
@@ -140,7 +140,7 @@
 {% endblock %}
 
 {% block footer %}
-{% if strategicAppChannel === true %}
+{% if isApp %}
 
 {% else %}
     {{ govukFooter({
--- a/src/components/mfa-reset-with-ipv/mfa-reset-with-ipv-controller.ts
+++ b/src/components/mfa-reset-with-ipv/mfa-reset-with-ipv-controller.ts
@@ -19,7 +19,7 @@ export function mfaResetWithIpvGet(
       );
     }
 
-    if (res.locals.strategicAppChannel === true) {
+    if (res.locals.isApp) {
       const redirectPath = await getNextPathAndUpdateJourney(
         req,
         req.path,
--- a/src/components/mfa-reset-with-ipv/tests/mfa-reset-with-ipv-integration.test.ts
+++ b/src/components/mfa-reset-with-ipv/tests/mfa-reset-with-ipv-integration.test.ts
@@ -58,26 +58,28 @@ describe("Mfa reset with ipv", () => {
       );
     });
 
-    it("should redirect to the new guidance page when using the strategic app", async () => {
-      const previousPath = PATH_NAMES.ENTER_AUTHENTICATOR_APP_CODE;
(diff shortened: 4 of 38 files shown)


src/middleware/tests/channel-middleware.test.ts
GHSwallow, line 1: There are quite a few tests which test specifically strategic app - eg `base-integration`, `enter-email-controller`, `mfa-reset-with-ipv-controller` and `mfa-reset-with-ipv-integration`. Perhaps worth making each of these also test `MOBILE` as you did in `account-not-found-controller`?

src/components/sign-in-or-create/tests/sign-in-or-create-controller.test.ts
```diff
@@ -32,18 +32,17 @@ describe("sign in or create controller", () => {
       expect(res.render).to.have.calledWith("sign-in-or-create/index.njk");
     });
 
-    describe("where the context is mobile", () => {
-      it("should render the mobile template", async () => {
-        res.locals.strategicAppChannel = true;
+    it("should render the mobile template when the channel is an app", async () => {
```
GHSwallow, line 35: nice refactor!

src/components/common/layout/base.njk
```diff
@@ -6,7 +6,7 @@
 
 {% from "../header/macro.njk" import strategicAppHeader %}
 
-{% if strategicAppChannel == true %}
+{% if isApp %}
```
GHSwallow, line 9: Have you seen this work, or do we need the comparison `isApp == true` like before?

src/components/mfa-reset-with-ipv/mfa-reset-with-ipv-controller.ts
```diff
@@ -19,7 +19,7 @@ export function mfaResetWithIpvGet(
       );
     }
 
-    if (res.locals.strategicAppChannel === true) {
+    if (res.locals.isApp) {
```
GHSwallow, line 22: love using a boolean that's named like a boolean so we don't have to use an equality to make that clear 👍

src/components/account-created/account-created-controller.ts
```diff
@@ -7,7 +7,6 @@ export function accountCreatedGet(req: Request, res: Response): void {
   res.render("account-created/index.njk", {
     serviceType,
     name,
```
GHSwallow, line 9: Do you think these are strictly necessary? Or if they are, then I basically wonder if there are other cases of us passing in properties incorrectly into res.render throughout the codebase. Would be good to look into quickly, and make a tech debt ticket to fix this if it's a theme and also raise team-wide how this interface works. It's misleading and also will render things wrong

test/unit/error-handler.test.ts
```diff
@@ -50,21 +50,5 @@ describe("Error handlers", () => {
         "common/errors/session-expired.njk"
       );
     });
-
-    it("should pass through the strategicAppChannel", () => {
```
GHSwallow, line 54: Agree that these tests don't really make loads of sense anymore. We'd essentially be unit testing how res.render works. Am I right in saying then, we don't have any test coverage for rendering different pages depending on the value of `isApp`? Not that this test even did that, but it did something.

src/components/security-code-error/security-code-error-controller.ts
```diff
@@ -59,7 +59,6 @@ export function securityCodeTriesExceededGet(
     ),
     isResendCodeRequest: req.query.isResendCodeRequest,
     isAccountCreationJourney: req.session.user?.isAccountCreationJourney,
```
GHSwallow, line 61: ah, here are two examples of this being done incorrectly (I think, based on what we discussed).

src/components/security-code-error/tests/security-code-error-controller.test.ts
```diff
@@ -77,8 +77,6 @@ describe("security code controller", () => {
             newCodeLink: params.newCodeLink,
             isResendCodeRequest: undefined,
             isAccountCreationJourney: params.isAccountCreationJourney,
-
-            strategicAppChannel: res.locals.strategicAppChannel,
```
GHSwallow, line 81: I think we could in theory, check that isApp is passed in, but as I say, feels a bit like we're just testing how res.render works.

GHSwallow (commented):
looking good! Just some musings

src/components/common/layout/base.njk
alhcomer: It's been working in testing just fine. This syntax is used elsewhere also

src/components/mfa-reset-with-ipv/tests/mfa-reset-with-ipv-integration.test.ts
```diff
@@ -59,25 +59,27 @@ describe("Mfa reset with ipv", () => {
...
-      const agent = supertest.agent(app);
-
-      await agent
-        .get(REQUEST_PATH)
-        .set("Cookie", "channel=strategic_app")
-        .expect(302)
-        .expect("Location", expectedGuidancePage);
-
-      await agent.get(expectedGuidancePage).expect(200);
+      for (const channel of [CHANNEL.STRATEGIC_APP, CHANNEL.MOBILE]) {
```
GHSwallow, line 62: CHANNEL. MOBILE should be swapped with GENERIC_APP.
Also, could these be in separate tests so we have two tests, one for each? Gives clarity on which fails so we get a more granular unit being tested.

GHSwallow (approved):

PR #2800, opened 2025-05-23
ATO-1669: Pass client_name claim to backend

cearl1:
## What

We would like to send the client name to the auth backend, so auth does not need to rely on the client registry to get this value. We are already sending the claim from orch to auth frontend, this PR is passing this claim onto the backend.

Deployed to authdev and tested a few journeys. All were successful

## Checklist

- [x] Performance analyst has been notified of the change. n/a
- [x] A UCD review has been performed. n/a
- [x] Any necessary changes to the [acceptance tests](https://github.com/govuk-one-login/authentication-acceptance-tests) have been made. n/a
- [x] Documentation has been updated to reflect these changes. n/a

### DIFF
--- a/src/components/authorize/authorize-service.ts
+++ b/src/components/authorize/authorize-service.ts
@@ -82,5 +82,6 @@ function createStartBody(startRequestParameters: StartRequestParameters) {
   body["scope"] = startRequestParameters.scope;
   body["redirect_uri"] = startRequestParameters.rp_redirect_uri;
   body["state"] = startRequestParameters.rp_state;
+  body["client_name"] = startRequestParameters.client_name;
   return body;
 }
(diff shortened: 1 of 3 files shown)


src/components/authorize/authorize-service.ts
```diff
@@ -82,5 +82,6 @@ function createStartBody(startRequestParameters: StartRequestParameters) {
   body["scope"] = startRequestParameters.scope;
   body["redirect_uri"] = startRequestParameters.rp_redirect_uri;
   body["state"] = startRequestParameters.rp_state;
+  body["client_name"] = startRequestParameters.client_name;
```
Ryan-Andrews99, line 85: Is this a required field on the JAR and if not should it be?

cearl1: Yep it's already required: 
https://github.com/govuk-one-login/authentication-frontend/blob/1423acfdc3cbb90efe03b8a24347b4eafd54b92d/src/components/authorize/claims-config.ts#L58-L65

Ryan-Andrews99 (approved):

PR #2893, opened 2025-07-22
BAU: Bump express to v5

alhcomer:
## What

- Bumps the npm-express-dependencies group with 3 updates in the / directory: [express](https://github.com/expressjs/express), [@types/express](https://github.com/DefinitelyTyped/DefinitelyTyped/tree/HEAD/types/express) and [express-validator](https://github.com/express-validator/express-validator).

- See change descriptions in commit messages

## How to review

2. Run through journey, go to create account, submit empty email, see validation error
3. Deploy to an environment and run acceptance tests against branch

## Checklist


src/app.ts
```diff
@@ -297,7 +296,7 @@ async function createApp(): Promise<express.Application> {
   app.use(channelMiddleware);
   app.use(environmentBannerMiddleware);
   app.use(getSessionIdMiddleware);
-  app.post("*", sanitizeRequestMiddleware);
+  app.post("/*splat", sanitizeRequestMiddleware);
```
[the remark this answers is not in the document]
alhcomer: splat is the wildcard param name express mention to use in the [v5 migration guide](https://expressjs.com/en/guide/migrating-5.html#:~:text=node%20index.js-,Changed,behavior%20of%20parameters%20%3A%2C%20use%20/*splat%20instead%20of%20/*,-//%20v4%0Aapp.)

src/middleware/form-validation-middleware.ts
```diff
@@ -1,18 +1,31 @@
...
+      };
+    case "alternative":
+    case "alternative_grouped":
+      return error.msg;
+    case "unknown_fields": {
+      const fields = error.fields.map((field) => field.path).join(", ");
+      return `Unknown fields found, please remove them: ${fields}`;
+    }
+    default:
+      throw new Error(`Not a known express-validator error: ${error}`);
```
BeckaL: This interpolated error will get serialised as [object Object]

src/middleware/form-validation-middleware.ts
```diff
@@ -1,18 +1,31 @@
...
+) => {
+  switch (error.type) {
+    case "field":
+      return {
+        text: error.msg,
+        href: `#${error.path}`,
+      };
+    case "alternative":
+    case "alternative_grouped":
+      return error.msg;
```
BeckaL: We have an issue here that this type is now different to the type that renderBadRequest expects.

As far as I understand it, these errors can't currently be thrown in our codebase since they're for when oneOf validation is used, and we don't use it currently in this codebase. An alternative here might be to throw an error for these cases, and force it to be handled explicitly in the associated change PR if we ever start using these validations. I think I'd prefer this rather than shaping the return type here to something expected, since likely if we got to the point of implementing this, we'd want to think about whether the types here work or not / it'd be easier to work from a concrete example.

src/middleware/form-validation-middleware.ts
```diff
@@ -1,18 +1,31 @@
...
+    case "unknown_fields": {
+      const fields = error.fields.map((field) => field.path).join(", ");
+      return `Unknown fields found, please remove them: ${fields}`;
```
BeckaL: Same issue here. From some reading, this error case is for if we were using [checkExact](https://express-validator.github.io/docs/api/check-exact/), but we don't currently. Even if we were, it would potentially make more sense to throw an error.

BeckaL (approved):

PR #2952, opened 2025-08-14
BAU: Add dark mode support to journey map

ben-fernandes-sw:
## What

Add styling to make journey map respond to system light/dark mode.

## How to review

1. Code Review
2. Run locally and see that the journey map responds to your system light/dark mode

### Screenshots

| Light mode | Dark mode |
|--------|--------|
| image | image |


Joe-Edwards-GDS (approved):
Happy to approve as an interim option, though looks like you could probably improve a lot of the label styles further.

There is theming configuration available in mermaid (https://mermaid.js.org/config/theming.html#theme-configuration), which is probably a more sustainable approach, but would require some JS to pick up dark mode and apply the theme

PR #2955, opened 2025-08-15
Aut 4440/template viewer

Joe-Edwards-GDS:
## What

Add template viewer to auth frontend in non-prod environments. Currently only includes main (non optional) journey states, but can be easily extended.

This allows you to see a list of pages in the app:

image

Selecting one allows you to preview that page outside of a journey:

image

The journey map is also updated so that double-clicking on a page state (in green) opens up a preview of that page:

image

## How to review

1. Code Review
1. Test locally with `yarn dev` and visiting both https://localhost:3000/templates and https://localhost:3000/journey-map
1. [Optional] Deploy to a dev environment and test there

### DIFF
--- a/src/components/templates/pages.ts
+++ b/src/components/templates/pages.ts
@@ -1,0 +1,83 @@
+import { PATH_NAMES } from "../../app.constants.js";
+
+interface Page {
+  template: string;
+  options?: Record<string, unknown>;
+}
+
+interface PageVariant extends Page {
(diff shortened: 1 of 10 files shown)


src/components/templates/pages.ts
Joe-Edwards-GDS, line 1: This reminds me a lot of something like [storybook](https://storybook.js.org/)

PR #2964, opened 2025-08-22
AUT-4686: Common journey states

Joe-Edwards-GDS:
## What

- Added common state for 'sign-in end', abstracting away the need to distinguish T&Cs accepted in each final auth step
- Added common state for 'password verified', ensuring both variants map the same paths
- Some styling tweaks to highlighted states and labels

image

Although this adds more states to the map, it reduces the total number of elements as it eliminates a bunch of parallel state transitions.

A more sophisticated approach here could instead use a 'sub-journey', completely abstracting the common states.

## How to review

1. Code Review
1. Marvel at journey map locally
1. Ensure integration tests pass

### DIFF
--- a/src/components/common/state-machine/state-machine.ts
+++ b/src/components/common/state-machine/state-machine.ts
@@ -6,6 +6,11 @@ import {
   PATH_NAMES,
 } from "../../../app.constants.js";
 
+const INTERMEDIATE_STATES = {
+  PASSWORD_VERIFIED: "password-verified",
+  SIGN_IN_END: "sign-in-end",
+};
+
 const USER_JOURNEY_EVENTS = {
   AUTHENTICATED: "AUTHENTICATED",
   CREDENTIALS_VALIDATED: "CREDENTIALS_VALIDATED",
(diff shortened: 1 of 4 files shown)


src/components/common/state-machine/state-machine.ts
```diff
@@ -685,7 +668,7 @@ const authStateMachine = createMachine<AuthStateContext>(
   },
   {
     guards: {
-      isLatestTermsAndConditionsAccepted: (context) =>
+      needsLatestTermsAndConditions: (context) =>
```
Joe-Edwards-GDS, line 671: Previously `isLatestTermsAndConditionsAccepted` was the inverse of `context.isLatestTermsAndConditionsAccepted` 🙃

src/components/common/state-machine/state-machine.ts
```diff
@@ -563,6 +537,15 @@ const authStateMachine = createMachine<AuthStateContext>(
           ],
         },
       },
+      [INTERMEDIATE_STATES.SIGN_IN_END]: {
+        always: [
```
Joe-Edwards-GDS, line 541: This is the trick - an `always` transition will immediately resolve the next state (but can still use conditions)

PR #2969, opened 2025-09-02
BAU: State machine auto context

Joe-Edwards-GDS:
## What

Compute state machine context from current session instead of passing in bespoke parameters at each call-site. This should make it easier to reason about the state machine and use existing conditions elsewhere.

Also:
- Refine logging to make it easier to analyse journeys
- Remove some impossible transitions

New journey map:

image

## How to review

1. Code Review
1. Ensure all acceptance tests still pass

## Related PRs

https://github.com/govuk-one-login/authentication-frontend/pull/2962

### DIFF
--- a/src/components/enter-password/tests/enter-password-integration.test.ts
+++ b/src/components/enter-password/tests/enter-password-integration.test.ts
@@ -13,6 +13,7 @@ import type { NextFunction, Request, Response } from "express";
 import { getPermittedJourneyForPath } from "../../../../test/helpers/session-helper.js";
 import esmock from "esmock";
 import { buildMfaMethods } from "../../../../test/helpers/mfa-helper.js";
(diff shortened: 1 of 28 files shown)


src/components/enter-password/tests/enter-password-integration.test.ts
```diff
@@ -120,17 +121,21 @@ describe("Integration::enter password", () => {
       .expect(400);
   });
 
-  it("should redirect to /auth-code when password is correct (VTR Cm)", async () => {
+  it("should redirect to /auth-code when password is correct (VTR Cl)", async () => {
```
Joe-Edwards-GDS, line 124: Not sure if this was a typo in the test name or the implementation - but this should be a bit more robust

PR #3241, opened 2026-03-24
DFC-1100: Add GA4 isPageDataSensitive flags and contentIds across pages

DavidMcMichael-GDS:
## What

Updating content ID's to the new ones provided by the data team and setting the isPageDataSensitive on each page in order to allow non PII data through to analytics 

## Checklist


src/components/contact-us/contact-us-controller.ts
```diff
@@ -123,6 +123,168 @@ const somethingElseSubThemeToPageTitle = {
     "pages.contactUsQuestions.govUKLoginAndIdAppsSomethingElse.title",
 };
 
+const ga4DataSetterContactUsQuestions: {
```
mattvot: Could these content ids move to `contentIds.ts` please

src/components/common/layout/base.njk
```diff
@@ -191,7 +195,8 @@
                         uaContainerId: "not used" // We don't use UA anymore, but it's a required param
                     }, {
                         enableGa4Tracking: {{isGa4Enabled}},
-                        cookieDomain: "{{analyticsCookieDomain}}"
+                        cookieDomain: "{{analyticsCookieDomain}}",
+                        isPageDataSensitive: {{isPageDataSensitive}},
```
mattvot, line 199: We have a `getAnalyticsPropertiesMiddleware` that runs before the routes these templates are rendered. This will fetch properties from res.locals if set, else fallback to templates. 

We use this for existing contentIds and taxonomies here https://github.com/govuk-one-login/authentication-frontend/blob/main/src/components/ga4-opl/template.njk

Can we use a similar approach here, where we assume all pages are not sensitive, unless overwritten in templates as you have done in our existing templates (which would mean we can remove all `false` setting, leaving only `true`?

src/components/contact-us/contact-us-controller.ts
```diff
@@ -400,6 +562,7 @@ export function furtherInformationGet(req: Request, res: Response): void {
     radioButtons: getThemeRadioButtonsFromStructure(themeStructure.subThemes),
     title: getTitleKeyFromTheme(themeStructure),
     header: getHeaderKeyFromTheme(themeStructure),
+    contentId: getGA4DataVariablesContactUsQuestions(req).contentId,
```
mattvot, line 565: In theory this should not be needed so long as `getContentId` of `contentIds.ts` is returning the content ids of the path. That way no content id changes should be needed in this file

src/components/how-do-you-want-security-codes/index.njk
```diff
@@ -57,6 +57,7 @@
     </p>
 {% endif %}
 
+{% set isPageDataSensitive = true %}
```
mattvot, line 60: Is this true because it render the last few digits of the number? If so, there are a few other pages that do this I could suggest

mattvot (changes requested):
Thanks for making these changes. I've suggested some structural changes with how we define contentIds and how we access them.

src/components/account-created/index.njk
```diff
@@ -38,6 +38,7 @@
 {% endif %}
```
mattvot, line 38: Just leaving a comment here to track resolution. Please can you update the commit message with a description of why this change is being made please (ideally breaking down the commit into several to tell a story of the change). See https://team-manual.account.gov.uk/teams/authentication/working-with-git#commit-messages. 

Many thanks

mattvot (changes requested):

src/components/common/layout/base.njk
DavidMcMichael-GDS: Would be cautious about this one as setting the default as false would be a lot more risky if a new page were to be added and sensitive data ended up being captured

src/components/contact-us/contact-us-controller.ts
DavidMcMichael-GDS: The contact-us content IDs are based on query param combinations (theme/subtheme), not req.path, so getContentId can't resolve them. They need to be passed explicitly from the controller where the query params are available.

src/components/how-do-you-want-security-codes/index.njk
DavidMcMichael-GDS: Im unsure on this. The data team decide which pages are set as sensitive so one to raise with them

dbes-gds:
closed in favour of #3480

PR #3331, opened 2026-04-22
AUT-4800: Collect information on if browser supports passkeys

ChristianSoftwire:
## What

- Add a new javascript module to check if the browser supports webauthn. If it does, set the value of a hidden input on the template to be "true". Handle this value in the enterEmailPost controller and set it as a boolean in the session
- Only allow `shouldPromptToRegisterPasskey` and `shouldPromptToSignInWithPasskey` if this value is `true`

## How to review

1. Code review
2. Test in a deployed dev env

### DIFF
--- a/src/assets/javascript/check-passkeys-supported.js
+++ b/src/assets/javascript/check-passkeys-supported.js
@@ -1,0 +1,9 @@
+/* global SimpleWebAuthnBrowser */
+document.addEventListener("DOMContentLoaded", () => {
+  if (SimpleWebAuthnBrowser.browserSupportsWebAuthn()) {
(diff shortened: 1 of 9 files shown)


src/assets/javascript/check-passkeys-supported.js
```diff
@@ -0,0 +1,9 @@
+/* global SimpleWebAuthnBrowser */
```
ChristianSoftwire, line 1: `simplewebauthn-browser.min.js` is loaded via a <script> in `base.njk` so is available globally, but sonar can't pick that up so was throwing a failure. Added this in to remove that

BeckaL (approved):
LGTM

PR #3355, opened 2026-05-01
BAU: Change allow listed client ID

alhcomer:
## What

The orch stub previously had the same client ID as the RP stub. This would have broken the new allow listing, so the rpClientId sent by the orch stub has been updated to be unique and is allow listed here.

## How to review

1. Code Review

## Checklist


ChristianSoftwire (approved):
LGTM

PR #3356, opened 2026-05-01
AUT-5068: Track history from enter email to enter password

ChristianSoftwire:
## What

- Add goBackHistory to session and res.locals so it can be exposed for templates
- Make transition from enter-email to enter-password reversible
- Add the previousPath to goBackHistory if passkeys are enabled and the transition was reversible

## How to review

1. Code review
2. Deploy to dev env, run through a journey and check that the `goBackHistory` is what you expect it to be (ie enter-email once we get to enter-password and then it should stay as that as the journey progresses)

### DIFF
--- a/src/components/common/state-machine/state-machine-executor.ts
+++ b/src/components/common/state-machine/state-machine-executor.ts
@@ -10,6 +10,7 @@ import {
   shouldPromptToSignInWithPasskey,
 } from "../../../utils/passkeys-helper.js";
 import { needsForcedMFAReset } from "../../../utils/request.js";
+import { getGoBackHistoryForTransition } from "./go-back-history.js";
 
 export async function getNextPathAndUpdateJourney(
   req: Request,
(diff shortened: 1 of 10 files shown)


src/components/common/state-machine/state-machine-executor.ts
```diff
@@ -39,13 +40,25 @@ export async function getNextPathAndUpdateJourney(
...
+  const isReversibleTransition =
+    nextState.transitions[0]?.meta?.reversible ?? false;
+  const passkeysEnabled =
+    res.locals.supportPasskeyRegistration || res.locals.supportPasskeyUsage;
+  const goBackHistory = buildGoBackHistoryForTransition(
+    req.session.user?.journey?.goBackHistory ?? [],
+    passkeysEnabled,
+    currentState,
+    isReversibleTransition
+  );
```
ben-fernandes-sw: Could extract this to its own function

ben-fernandes-sw (approved):

ben-fernandes-sw (approved):

PR #3365, opened 2026-05-06
AUT-5324: Add environment variable mapping for client allow list

alhcomer:
## What

The allow-list was coming through as undefined (so no clients allowed) as there was no environment variable mapping in cloudformation. This change makes sure environment variables are picked up in ECS

## How to review

1. Code Review

## Checklist

### DIFF
--- a/cloudformation/deploy/template.yaml
+++ b/cloudformation/deploy/template.yaml
@@ -492,6 +492,7 @@ Mappings:
       SupportNewInternationalSms: "No"
       EnableDwpKbvContactFormChanges: "Yes"
       SupportSingleFactorAccountDeletion: "Yes"
+      PasskeyPromptClientAllowList: ""
     dev:
       DeployServiceDownPage: "No"
       dynatraceSecretArn: arn:aws:secretsmanager:eu-west-2:216552277552:secret:DynatraceNonProductionVariables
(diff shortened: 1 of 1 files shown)


cloudformation/deploy/template.yaml
```diff
@@ -492,6 +492,7 @@ Mappings:
       SupportNewInternationalSms: "No"
       EnableDwpKbvContactFormChanges: "Yes"
       SupportSingleFactorAccountDeletion: "Yes"
+      PasskeyPromptClientAllowList: ""
```
dbes-gds, line 495: does it let everyone through if empty?

alhcomer: shows nobody passkey prompts if the allow list is empty

see:

```
export function getPasskeyPromptClientAllowList(): string[] {
  const rawList = process.env.PASSKEY_PROMPT_CLIENT_ALLOW_LIST;
  if (rawList) {
    return rawList.split(",").map((id) => id.trim());
  }
  return [];
}
```

and: 
```
function isPromptableRPClientID(rpClientId: string) {
  return getPasskeyPromptClientAllowList().includes(rpClientId);
}
```

dbes-gds (approved):

PR #3367, opened 2026-05-06
AUT-5324: set amc cookie on .account.gov.uk domain

alhcomer:
## What

- Required for AMC to get the cookie when integrating
- Without domain specified, res.cookie() sets cookie as host only
- getGoogleAnalyticsAndDynatraceCookieDomain() is the existing way to get .account.gov.uk domain if SERVICE_DOMAIN is not "localhost", which it isn't in environments
- Also added `config.test.ts` file
## How to review

1. Code Review

## Checklist

### DIFF
--- a/src/components/create-passkey/create-passkey-controller.ts
+++ b/src/components/create-passkey/create-passkey-controller.ts
@@ -7,6 +7,7 @@ import type { AmcAuthorizeInterface } from "../amc-service/types.js";
 import { amcAuthorizeService } from "../amc-service/amc-authorize-service.js";
 import type { ExpressRouteFunc } from "../../types.js";
 import { BadRequestError } from "../../utils/error.js";
+import { getGoogleAnalyticsAndDynatraceCookieDomain } from "../../config.js";
 
 const TEMPLATE_NAME = "create-passkey/index.njk";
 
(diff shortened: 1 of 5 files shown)


src/components/create-passkey/tests/create-passkey-controller.test.ts
```diff
@@ -86,6 +86,7 @@ describe("create passkey controller", () => {
       expect(res.cookie).to.have.been.calledWith("amc", AMC_COOKIE, {
         secure: true,
         httpOnly: true,
+        domain: "localhost",
```
ChristianSoftwire, line 89: Would be good to have a test to also check that this gets set correctly in higher envs

ChristianSoftwire (commented):
Just a small comment about a test

src/components/create-passkey/tests/create-passkey-controller.test.ts
alhcomer: I'm a little against testing that in a controller test as the functionality is separate, so have added a `config.test.ts` file

src/components/create-passkey/create-passkey-controller.ts
```diff
@@ -73,6 +74,7 @@ async function handleCreatePasskey(
   res.cookie("amc", result.data.amcCookie, {
     secure: true,
     httpOnly: true,
+    domain: getGoogleAnalyticsAndDynatraceCookieDomain(),
```
BeckaL: I think we should rename this method - maybe to getAccountDomain?

alhcomer: Updated in new commit, thanks :)

BeckaL (approved):

PR #3377, opened 2026-05-07
AUT-5067: Use goBackHistory when going from enter-password to enter-email

ChristianSoftwire:
## What

- Update `allow-user-journey-middleware` to allow transitions back validating against the goBackHistory. This extra validation is only true if passkeys are enabled and the user has come from an approved path in the allowList
- Add `go-back-history-middleware` to handle popping from the goBackHistory when a back is registered from an allowed path in the allowList. Add this new middleware to the enter-email route so that goBackHistory can be used to go back from there
- Update the `enter-password` nunjucks template to set the href to be the last element of goBackHistory if passkeys are enabled and it's not empty

## How to review

1. Code review (I have put quite abit of detail in the commit messages so would urge reading those)
2. Run locally or deploy to dev env and manually test that going back and forward works. In particular, pay attention to the sign-in-or-create -> enter-email -> enter-password -> forgot password. And then go back and check it all works as expected. Check logs and ensure goBackHistory is being updated as you expect.

### DIFF
--- a/src/components/common/state-machine/go-back-history.ts
+++ b/src/components/common/state-machine/go-back-history.ts
@@ -1,5 +1,8 @@
 import type { Request, Response } from "express";
 import type { AuthState } from "./state-machine.js";
+import { PATH_NAMES } from "../../../app.constants.js";
+
+export const goBackHistoryAllowList = [PATH_NAMES.ENTER_PASSWORD];
 
 export function getGoBackHistoryForTransition(
   req: Request,
@@ -32,4 +35,16 @@ export function buildGoBackHistoryForTransition(
   }
 
   return [...currentGoBackHistory, previousState];
+}
+
+export function isBackTransition(
+  goBackHistory: string[],
+  currentPath: string
+): boolean {
+  if (goBackHistory.length === 0) {
+    return false;
+  }
+
+  const lastPath = goBackHistory[goBackHistory.length - 1];
+  return lastPath === currentPath;
 }
--- a/src/components/sign-in-or-create/sign-in-or-create-controller.ts
+++ b/src/components/sign-in-or-create/sign-in-or-create-controller.ts
@@ -10,6 +10,10 @@ export async function signInOrCreateGet(
   req.session.user.isAccountCreationJourney = false;
   req.session.user.isPasswordResetJourney = false;
   req.session.user.isSignInJourney = false;
+  // Reset goBackHistory as users can navigate back here from enter-password via optionalPaths, making the history stale
+  if (req.session.user.journey) {
+    req.session.user.journey.goBackHistory = [];
+  }
   if (req.query.redirectPost) {
     return await signInOrCreatePost(req, res);
   }
--- a/src/middleware/allow-user-journey-middleware.ts
+++ b/src/middleware/allow-user-journey-middleware.ts
@@ -1,12 +1,27 @@
 import type { NextFunction, Request, Response } from "express";
 import { authStateMachine } from "../components/common/state-machine/state-machine.js";
 import { saveSessionState } from "../components/common/constants.js";
-export function transitionForbidden(req: Request): boolean {
+import {
+  goBackHistoryAllowList,
+  isBackTransition,
+} from "../components/common/state-machine/go-back-history.js";
+
+export function transitionForbidden(
+  req: Request,
+  passkeysEnabled?: boolean
+): boolean {
   const nextPath = req.session.user.journey.nextPath;
-  return (
-    nextPath !== req.path &&
-    !req.session.user.journey.optionalPaths.includes(req.path)
+
+  const isOnNextPath = nextPath === req.path;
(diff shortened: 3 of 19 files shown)


alhcomer (approved):
really nice work. tested locally and working as expected

alhcomer (dismissed):

src/components/common/state-machine/go-back-history.ts
```diff
@@ -33,3 +36,15 @@ export function buildGoBackHistoryForTransition(
...
+export function isBackTransition(
+  goBackHistory: string[],
+  currentPath: string
+): boolean {
+  if (goBackHistory.length === 0) {
+    return false;
+  }
+
+  const lastPath = goBackHistory[goBackHistory.length - 1];
+  return lastPath === currentPath;
```
ben-fernandes-sw, line 49: You're right, naming is tricky.

Maybe
```
return lastPathOfHistory === pathNavigatingTo
```

src/components/sign-in-or-create/sign-in-or-create-controller.ts
```diff
@@ -10,6 +10,9 @@ export async function signInOrCreateGet(
   req.session.user.isAccountCreationJourney = false;
   req.session.user.isPasswordResetJourney = false;
   req.session.user.isSignInJourney = false;
+  if (req.session.user.journey) {
+    req.session.user.journey.goBackHistory = [];
```
ben-fernandes-sw, line 15: Hmm, it's not always guaranteed that this is the first page a user will visit (I'm assuming that's why you've set it to an empty array here). They may be on a reauth journey which would skip over sign-in-or-create altogether and go straight to enter-email.

src/middleware/allow-user-journey-middleware.ts
```diff
@@ -1,12 +1,27 @@
...
-    nextPath !== req.path &&
-    !req.session.user.journey.optionalPaths.includes(req.path)
+
+  const isOnNextPath = nextPath === req.path;
+  const isOnOptionalPath = req.session.user.journey.optionalPaths.includes(
+    req.path
   );
+  const isGoingBack =
+    passkeysEnabled &&
+    goBackHistoryAllowList.includes(nextPath) &&
```
ben-fernandes-sw, line 21: I'm wondering if this would be better enforced by a state machine check of some form. Either
* metadata on the enter-password page state that says you're allowed to go back from it, or
* go to stateMachine[req.path], filter all the transitions down to ones that are reversible, check if any of those include nextPath

Just so that we keep as much of the routing config in the state machine as possible.

ben-fernandes-sw: Happy to leave that as a follow-up action

src/middleware/go-back-history-middleware.ts
```diff
@@ -0,0 +1,37 @@
...
+  const goBackHistory = req.session.user.journey.goBackHistory ?? [];
+  const lastPathInGoBackHistory = goBackHistory[goBackHistory.length - 1];
+
+  if (isBackTransition(goBackHistory, req.path)) {
+    if (!goBackHistoryAllowList.includes(previousPathInSession)) {
+      req.log.warn(
+        `User tried to use goBackHistory from invalid path ${previousPathInSession} in session and goBackHistory ${goBackHistory}`
+      );
+      return res.redirect(previousPathInSession);
+    }
```
ben-fernandes-sw: Does this not duplicate the changes to the allow-user-journey-middleware?

src/middleware/go-back-history-middleware.ts
```diff
@@ -0,0 +1,37 @@
+import type { NextFunction, Request, Response } from "express";
+import { saveSessionState } from "../components/common/constants.js";
+import { authStateMachine } from "../components/common/state-machine/state-machine.js";
+import {
+  goBackHistoryAllowList,
+  isBackTransition,
+} from "../components/common/state-machine/go-back-history.js";
+
+export async function goBackHistoryMiddleware(
```
ben-fernandes-sw: Maybe `goBackMiddleware` instead. I think the primary purpose of this middleware is to handle what happens when you go back. Updating the history array is just part of that.

ben-fernandes-sw (commented):
I'd appreciate a walkthrough of this PR. I think I've understood it, but want to double check.

src/components/sign-in-or-create/sign-in-or-create-controller.ts
ChristianSoftwire: TODO: Add a comment explaining that this is for the multiback from enter-password to fix the journey

src/middleware/allow-user-journey-middleware.ts
ChristianSoftwire: Agreed to leave as is for now and then change in a follow up ticket

ben-fernandes-sw (approved):

PR #3397, opened 2026-05-18
AUT-4989: Add new account exists page

BeckaL:
Adds a new page for the create account journey, when an account already exists for that email, and that account has an active passkey.

This is because when an email is entered in a create account journey that already exists, we prompt the user to sign in instead of creating the account. And when they sign in, we need to know whether or not to route them via the passkey sign in.

A user will only be routed to the new page if all of the following apply:

* They have a passkey
* They passkeys usage feature flag is turned on
* They have javascript enabled
* Their browser supports passkey usage

In environments where the feature flag is turned off, users will not be routed to the new page, and manually navigating to the new page will result in a 404 (consistent with other urls that don't exist).

## How to review

* Code review commit by commit
* Run locally, and test this out. I hacked this by manually setting the hasActivePasskey value on the session. You can test: 
   - feature flag on / off
   - hasActivePasskey true / false
   - turning off javascript in your browser 

A related change will be made in the acceptance test repo. However, since we do not currently have supportPasskeyUsage turned on in any environment, there is no dependency between this PR and the acceptance tests - it can be merged as it will not change any journeys in any deployed environment.
  
## Before merge

[x] UCD review has been performed


src/components/account-exists-with-passkey/account-exists-with-passkey-routes.ts
```diff
@@ -0,0 +1,16 @@
+import { PATH_NAMES } from "../../app.constants.js";
+import * as express from "express";
```
alhcomer: should this be a wildcard?

BeckaL: Good point!

alhcomer (dismissed):

src/app.constants.ts
```diff
@@ -82,6 +82,7 @@ export const PATH_NAMES = {
   CREATE_PASSKEY_CALLBACK: "/create-passkey-callback",
   SESSION_ENDED: "/session-ended",
   PASSKEY_CREATED: "/passkey-created",
+  ACCOUNT_EXISTS_WITH_PASSKEY: "/account-exists",
```
ben-fernandes-sw, line 85: Thoughts on just calling this `ACCOUNT_EXISTS` to match our other paths?
I'm conflicted, given that you'd only end up here on a passkey journey, but the page content is definitely not passkey specific.

src/components/account-exists-with-passkey/account-exists-with-passkey-routes.ts
```diff
@@ -1,9 +1,16 @@
...
+import { allowUserJourneyMiddleware } from "../../middleware/allow-user-journey-middleware.js";
 
 const router = express.Router();
 
-router.get(PATH_NAMES.ACCOUNT_EXISTS_WITH_PASSKEY, accountExistsWithPasskey);
+router.get(
+  PATH_NAMES.ACCOUNT_EXISTS_WITH_PASSKEY,
+  validateSessionMiddleware,
+  allowUserJourneyMiddleware,
+  accountExistsWithPasskey
```
ben-fernandes-sw: Suffix with `Get` to match the pattern of our other controller functions

src/components/enter-email/enter-email-controller.ts
```diff
@@ -189,6 +189,10 @@ export function enterEmailCreatePost(
 
     req.session.user.email = email;
 
+    const { browserSupportsWebAuthn } = req.body;
+    req.session.user.browserSupportsWebAuthn =
+      browserSupportsWebAuthn === "true";
```
ben-fernandes-sw: Nit pick - The beginning few lines of the `enterEmailPost` function look almost identical, but in a slightly different order. Could we change this function to match that pattern.

ben-fernandes-sw (dismissed):
Looks good. Just a few nit-pick comments.

src/app.constants.ts
BeckaL: Yeah I also feel conflicted about this and went a bit back and forth. Thoughts are that the url is generic, but internally currently this is specific to a passkey, so I wanted it to be obvious when we're using it in code that it's (currently) a passkey specific route. So maybe we rename it if it gets used more generically?

src/components/enter-email/enter-email-controller.ts
BeckaL: Ah good spot, sorted

ben-fernandes-sw (approved):

PR #3404, opened 2026-05-20
BAU: Journey map reversible transitions

ben-fernandes-sw:
## What

Now that 'reversible' transitions exist in the state machine (as part of the back button work), it made sense to update the journey map to reflect.

### Arrow ends

A system for arrows in the journey map:
* `-->` Means the transition is final. There is no going back.
* `--o` Means the transition is reversible. It's possible to go back

I did consider having double-ended `<-->` arrows, but that doesn't display the primary direction of travel, so settled on having a slightly different end.

### Highlight colour

Now when you click on a state that has a reversible transition in or out from it, the transition is highlighted yellow. Given that you can go both ways along these arrows, it doesn't make sense for them to be blue (where you've come from) or yellow (where you're going to), as they can be both.

### Screenshot
Showing the `/enter-password` state selected
image

## How to review

1. Code Review
2. Run locally


sw-mattw (approved):
Nice, thanks

PR #3406, opened 2026-05-22
BAU: Allow the script to create passkeys for localhost

ChristianSoftwire:
## What

If you want to create a passkey for local running and then add the output to the local-authenticator table our generate webauthn script needs to support localhost as the rp ID.

## How to review

1. Code review

### DIFF
--- a/scripts/generate_webauthn/generate_webauthn_script.py
+++ b/scripts/generate_webauthn/generate_webauthn_script.py
@@ -1,6 +1,6 @@
 # --- CONFIGURATION: Update these values ---
 ENV_NAME = "Dev"
-ENV_URL = "dev"  # e.g., 'build' for build.account.gov.uk
+ENV_URL = "dev.account.gov.uk"  # e.g., 'build.account.gov.uk' for build (see README for examples)
 USER_PUB_SUB_ID = "xxx"
 USER_EMAIL = "xxx@xxx.xxx"
 CHALLENGE = "6EP1s53M5sF7XL_G2RwbY-AbJoSNiiCIVb9DrNNITnM"  # Base64URL encoded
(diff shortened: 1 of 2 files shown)


scripts/generate_webauthn/generate_webauthn_script.py
```diff
@@ -43,10 +43,18 @@
...
+            return "localhost"
+        }} else {{
+            return envUrl + ".account.gov.uk"
+        }}
+    }}
 
     const createOptions = {{
         challenge: bufferFromBase64("{CHALLENGE}"),
-        rp: {{ name: "GOV.UK One Login ({ENV_NAME})", id: "{ENV_URL}.account.gov.uk" }},
+        rp: {{ name: "GOV.UK One Login ({ENV_NAME})", id: getRpIdUrl("{ENV_URL}") }},
```
ben-fernandes-sw: What do you think about `ENV_URL` just being the full URL?

i.e. you need to define `build.account.gov.uk` fully yourself, and we just give some examples of what they could be

scripts/generate_webauthn/generate_webauthn_script.py
```diff
@@ -43,10 +43,18 @@
             aaguid: authData.slice(37, 53)
         }};
     }};
+    
+    const getRpIdUrl = (envUrl) => {{
+        if (envUrl === "local") {{
+            return "localhost"
```
ben-fernandes-sw: Doesn't it need to include the port, or is it happy with just `localhost`?

ChristianSoftwire: You can't have a port for the rp ID. But `localhost` works fine

PR #3440, opened 2026-06-04
AUT-4990: Build /cannot-sign-in-passkey page

ben-fernandes-sw:
## What

Builds:
* The error page
* Its validation
* Routing to it via the state machine from when passkey usage fails
* The associated unit/integration tests

Does not build:
* The conditional text on the radio button ("sign in with password" vs "sign in with password + mfa")
  * The data is not easily available until you enter your password, so we won't have it at point of rendering the error page. The template _has_ been built to handle both cases, so we _can_ provide screenshots, but the pages won't always be rendered correctly during a real journey. We'll pick up the conditional content in AUT-5054 (or in a new standalone ticket).
* The onward journeys from this error page
  * Currently, submitting the form just reloads the page. There are separate tickets (AUT-5000 and AUT-5001) to implement the onward functionality.

## How to review

1. Code Review
2. Test in `authdev1`
    * Set up a passkey for yourself using the `generate_webauthn_script.py` script
    * Try signing in with that passkey and intentionally fail using it (e.g. cancel out, let the prompt timeout)
    * Also try that disabling JS at various points in the journey will send you to the error page when you click "Continue"
    * Ensure you're redirected to the error page
    * The page should error out if no option is selected on the error page

## Checklist

- [ ] UCD review


src/components/cannot-sign-in-passkey/tests/cannot-sign-in-passkey-controller.test.ts
```diff
@@ -5,7 +5,10 @@ import type { RequestOutput, ResponseOutput } from "mock-req-res";
 import { mockResponse } from "mock-req-res";
 import { createMockRequest } from "../../../../test/helpers/mock-request-helper.js";
 import { PATH_NAMES } from "../../../app.constants.js";
-import { cannotSignInPasskeyGet } from "../cannot-sign-in-passkey-controller.js";
+import {
```
BeckaL: This commit is a bit confusing without an explanation in the commit message. Is the point here we're implementing a dummy post request until the next ticket is implemented, so we have a dummy implementation in here at the moment? 
My vote would be to implement this when it's needed for the next ticket, and so remove this commit from here. Will be easier to review in context.

scripts/generate_webauthn/generate_webauthn_script.py
```diff
@@ -19,7 +19,10 @@ def load_env(required_keys):
 
     missing = [k for k in required_keys if k not in env_vars]
     if missing:
-        print(f"Error: Missing required variables in .env: {', '.join(missing)}", file=sys.stderr)
```
BeckaL: If we end up amending anything in this PR, could we also put these linting fixes where they were originally required so we don't have this commit in the history (per https://team-manual.account.gov.uk/teams/authentication/working-with-git/#commits)

BeckaL (commented):
Looks sensible, tested and looks good to me, couple of small things

scripts/generate_webauthn/generate_webauthn_script.py
ben-fernandes-sw: I've gone through and merged them into where they came from.

For some reason, my pre-commit checker wasn't running 🤷 so it was more of a pain.

src/components/cannot-sign-in-passkey/tests/cannot-sign-in-passkey-controller.test.ts
ben-fernandes-sw: Good shout. I've dropped the new controller from this PR. Still requires building the post route to handle the form validation, but that's all for now.

BeckaL (approved):

PR #3414, opened 2026-05-28
AUT-5002: Successful passkey authentication

ChristianSoftwire:
## What

When a user clicks continue on the sign in with passkey page we want to sign them in with their passkey
How this works 
- On sign in with passkey get controller 
  - Call the StartPasskeyAssertionHandler to get the authenticationOptions (JSON blob returned from the java-webauthn-server package)
  - Render the page and pass the authenticationOptions in as a param
- When the user clicks continue on the page 
  - Call the startAuthentication method provided from browserWebauthn passing in the authenticationOptions from the StartAssertionHandler
  - Following a successful response, set a hidden inputs value to be the response
  - Call the form submit
- On sign in with passkey post controller
  - Send the response from the browserWebauthn to the FinishPasskeyAssertionHandler
  - This will check the browser handshake was done correctly and if so we receive a 200 on the frontend
  - Following a 200, move the user to `SIGN_IN_END`

## How to review

1. Code review
2. Deploy to a dev env, along with [this PR](https://github.com/govuk-one-login/authentication-api/pull/8386), and check you can sign in with a passkey (note, you will need to manually add a legitimate passkey to dynamo - see the `generate_webauthn_script` to see how to do that)

## Checklist

- [ ] A UCD review has been performed.

- [ ] Any necessary changes to the [acceptance tests](https://github.com/govuk-one-login/authentication-acceptance-tests) have been made.

## Related PRs

https://github.com/govuk-one-login/authentication-api/pull/8386

PR #3448, opened 2026-06-12
AUT-5001: Retry sign in with passkey (happy path)

ben-fernandes-sw:
## What

Adds the ability to retry using your passkey on the `/cannot-sign-in-passkey` screen.

Basically copies what was implemented in https://github.com/govuk-one-login/authentication-frontend/pull/3414 (the main `/sign-in-passkey` page) but is slightly different:
* refactors out some of the common code between the two screens
* some conditional logic based on which radio option was selected

## How to review

1. Code Review
1. Test - deployed to authdev3
    * The un-happy paths (failing to retry or using password instead) aren't implemented, so will error out

## Checklist

### DIFF
--- a/src/components/cannot-sign-in-passkey/index.njk
+++ b/src/components/cannot-sign-in-passkey/index.njk
@@ -52,3 +52,44 @@
     }) }}
   </form>
 {% endblock %}
+
+{% block scripts %}
+  <script {% if scriptNonce %} nonce="{{ scriptNonce }}"{%  endif %}>
+    const cannotSignInPasskeyForm = document.getElementById("cannotSignInPasskeyForm");
+    let isSubmitting = false
+
+    async function triggerWebAuthnFlow() {
+      try {
+        if (SimpleWebAuthnBrowser.browserSupportsWebAuthn()) {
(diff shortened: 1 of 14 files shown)


ben-fernandes-sw:
Sonar is complaining about code duplication, but a lot of that will be de-duped as part of AUT-5000, so I think it's safe to ignore.

src/components/cannot-sign-in-passkey/index.njk
```diff
@@ -52,3 +52,44 @@
...
+    async function handleSubmit(event) {
+      event.preventDefault();
+
+      if (isSubmitting) return
+      isSubmitting = true
+
+      const selectedOption = document.querySelector('input[name="cannot-sign-in-passkey-action"]:checked')?.value;
+
+      if (selectedOption === "retry-passkey") {
+        await triggerWebAuthnFlow();
```
ChristianSoftwire, line 85: Maybe not one for this ticket, but I can definitely see a case for extracting this method into somewhere reusable that can be used for sign-in-with-passkey and here

ChristianSoftwire (dismissed):
LGTM

src/components/cannot-sign-in-passkey/index.njk
ben-fernandes-sw: I'll follow this up on a future PR.

ChristianSoftwire (approved):

PR #3459, opened 2026-06-17
AUT-5522: Set mfaRequired in authorize

ChristianSoftwire:
## What

For passkeys, we need to know the `mfaRequired` flag earlier in the journey. This is so that we can display dynamic content on the cannot-sign-in-with-passkey page. However, we currently make this call in the `enter-password-controller` to the `/login` endpoint. 

This PR covers requesting the new `mfaRequired` flag in the authorize-controller and setting it in the frontend session. However, this PR doesn't remove any of the existing functionality in the enter-password-controller as that could affect existing sessions. Therefore, the plan is to first roll the changes out on this PR, then follow up with a subsequent PR to stop requesting the `mfaRequired` in the `enter-password-controller` and instead use the value directly from the session.

## How to review

1. Code review
2. Deploy to dev env and check that you don't see the added log and that journeys work as normal

### DIFF
--- a/src/components/enter-password/enter-password-controller.ts
+++ b/src/components/enter-password/enter-password-controller.ts
@@ -173,6 +173,12 @@ export function enterPasswordPost(
 
     const isPasswordChangeRequired = userLogin.data.passwordChangeRequired;
 
+    if (req.session.user.isMfaRequired !== userLogin.data.mfaRequired) {
+      req.log.info(
+        `isMfaRequired is ${req.session.user.isMfaRequired} but mfaRequired from /login is ${userLogin.data.mfaRequired}`
+      );
+    }
+
     req.session.user.mfaMethods = userLogin.data.mfaMethods;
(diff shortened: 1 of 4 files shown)


src/components/enter-password/enter-password-controller.ts
```diff
@@ -173,6 +173,10 @@ export function enterPasswordPost(
 
     const isPasswordChangeRequired = userLogin.data.passwordChangeRequired;
 
+    if (req.session.user.isMfaRequired !== userLogin.data.mfaRequired) {
+      req.log.info(`isMfaRequired is ${req.session.user.isMfaRequired} but mfaRequired from /login is ${userLogin.data.mfaRequired}`)
+    }
```
ChristianSoftwire, line 180: This is a temporary log to ensure that these values aren't different. I am planning to monitor the frontend logs for a while to make sure I can't see this log.

alhcomer (approved):

PR #3467, opened 2026-06-22
AUT-5522: Pass mfaRequired into the cannot-sign-in-passkey template to render the different content

ChristianSoftwire:
## What

Now that we are setting `isMfaRequired` in the `/authorize` we can pass this value into the cannot-sign-in-passkey template in order to render the dynamic content for whether to sign in with password or password and security code.

## How to review

1. Code review
2. Deploy to a dev env and go down a passkey sign in journey on both a 1fa and 2fa journey. When it comes to the sign in prompt, cancel and check that the content you see matches the authentication factor journey
  - Either "Sign in with your password"
  - "Sign in with your password and a security code"

## Checklist

### DIFF
--- a/src/components/cannot-sign-in-passkey/cannot-sign-in-passkey-controller.ts
+++ b/src/components/cannot-sign-in-passkey/cannot-sign-in-passkey-controller.ts
@@ -27,6 +27,7 @@ export function cannotSignInPasskeyGet(
 
     res.render("cannot-sign-in-passkey/index.njk", {
       authenticationOptions: JSON.stringify(authenticationOptions.publicKey),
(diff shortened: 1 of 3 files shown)


src/components/cannot-sign-in-passkey/cannot-sign-in-passkey-controller.ts
ben-fernandes-sw, line 1: You also need to make a change to the `postValidationLocals` in the `cannot-sign-in-passkey-validation.ts > validateBodyMiddleware()` call.

Otherwise, if you submit an invalid form, `is2FAJourney` will render as falsy.

ben-fernandes-sw (approved):

PR #3471, opened 2026-06-23
Aut 5503/remove route users to new ipv journey

BeckaL:
Removes two redundant variables from infrastructure. 

They are no longer used as of [this PR](https://github.com/govuk-one-login/authentication-frontend/pull/2749) but we missed removing them from the template at that time.

## How to review

Sense check my logic - are these variables used anywhere?


cloudformation/deploy/readme.md
alhcomer: Maybe we should delete "Now to enable MFA reset"

alhcomer (dismissed):

alhcomer (approved):

PR #3473, opened 2026-06-23
Aut 5529/suppress passkey setup for new accounts

BeckaL:
## What

Read the new field returned on user exists response which indicates if the backend knows of a reason to suppress the passkey registration prompt, and use it in the prompting rules.

Together with the backend PR, this means now that if a user has an account less than two hours old, we don't show them the register passkey prompt.

In future, more rules may be added into the backend.

## How to review

* Code review
* Run locally (or currently deployed to dev) with the backend PR. Check that all works as intended with an account without a passkey.

## Related PRs

[Backend PR](https://github.com/govuk-one-login/authentication-api/pull/8503) which needs to be merged first

An acceptance test PR will follow. Currently the acceptance tests which expect the passkey registration prompt will all still pass because when the user is setup in the database, we don't add a created at timestamp, and the backend treats this as no suppression required. So this can be merged before that.

### DIFF
--- a/src/components/enter-email/enter-email-controller.ts
+++ b/src/components/enter-email/enter-email-controller.ts
@@ -106,6 +106,8 @@ async function getExistingUserAndPopulateSessionData(
     { redactedPhoneNumber: result.data.phoneNumberLastThree }
   );
   req.session.user.hasActivePasskey = result.data.hasActivePasskey;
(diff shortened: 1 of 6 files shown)


ChristianSoftwire (approved):
LGTM

PR #3484, opened 2026-06-29
Aut 5531/suppress setup prompt in pw reset

BeckaL:
## What

Do not prompt a user to set up a passkey if they have been through any kind of password reset journey:

* voluntary password reset
* forced password reset via an account intervention
* forced password reset via having a common password

This is to reduce prompt fatigue when a user has gone through a more complicated sign in journey.

## Related PRs

[Related acceptance test change](https://github.com/govuk-one-login/authentication-acceptance-tests/pull/942) (this frontend PR can be merged first as the registration step is currently optional so will not fail if removed from these journeys, but this test PR prevents regression)


alhcomer (approved):
LGTM - was definitely sensible to introduce the new session item

PR #3516, opened 2026-07-03
AUT-5032: back button behaviour enter-password page to sign-in-passkey page

ChristianSoftwire:
## What

The `goBackHistory` has been added to the users frontend session already. We currently have the mechanism for tracking the users history during forward transitions and using stack to drive back navigation.

However, we have links on the site that do direct GET requests to the page in the a tag. These transitions are very difficult to track in the`goBackHistory` as they don't go via the state machine.

This PR introduces `/journey` which takes the "page the user was on" and the "event to go to the next page" as route params. It passes those into our state machine logic to get the next page the user should be on, adding the "page the user was on" to the `goBackHistory` so that we can track navigation via links in the stack.

See [this page](https://govukverify.atlassian.net/wiki/spaces/LO/pages/6574080144/Dynamic+Back+Behaviour) for more detail

- [x] Any necessary changes to the [acceptance tests](https://github.com/govuk-one-login/authentication-acceptance-tests) have been made.

- [x] Documentation has been updated to reflect these changes.

### DIFF
--- a/src/app.ts
+++ b/src/app.ts
@@ -113,6 +113,7 @@ import { passkeyCreatedRouter } from "./components/passkey-created/passkey-creat
 import { setGoBackHistoryMiddleware } from "./middleware/set-go-back-history-middleware.js";
 import { accountExistsWithPasskeyRouter } from "./components/account-exists-with-passkey/account-exists-with-passkey-routes.js";
 import { cannotSignInPasskeyRouter } from "./components/cannot-sign-in-passkey/cannot-sign-in-passkey-routes.js";
+import { journeyRouter } from "./components/journey/journey-routes.js";
 
 const directory_name = dirname(fileURLToPath(import.meta.url));
 
@@ -187,6 +188,7 @@ function registerRoutes(app: express.Application) {
     app.use(signInWithPasskeyRouter);
     app.use(accountExistsWithPasskeyRouter);
     app.use(cannotSignInPasskeyRouter);
+    app.use(journeyRouter);
   }
 
   // Development tools
--- a/src/components/common/state-machine/state-machine.ts
+++ b/src/components/common/state-machine/state-machine.ts
@@ -364,9 +364,18 @@ const authStateMachine = createMachine<AuthStateContext>(
               target: [INTERMEDIATE_STATES.SIGN_IN_END],
             },
           ],
+          [USER_JOURNEY_EVENTS.SIGN_IN_WITHOUT_PASSKEY]: [
+            {
+              target: [PATH_NAMES.ENTER_PASSWORD],
+              meta: { reversible: true },
+            },
+          ],
         },
         meta: {
           optionalPaths: [PATH_NAMES.ENTER_PASSWORD],
+          permittedJourneyRouteEvents: [
+            USER_JOURNEY_EVENTS.SIGN_IN_WITHOUT_PASSKEY,
+          ],
         },
       },
       [PATH_NAMES.CANNOT_SIGN_IN_PASSKEY]: {
--- a/src/components/common/state-machine/state-machine.ts
+++ b/src/components/common/state-machine/state-machine.ts
@@ -372,7 +372,6 @@ const authStateMachine = createMachine<AuthStateContext>(
           ],
         },
         meta: {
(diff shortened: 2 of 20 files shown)


src/components/common/state-machine/state-machine.ts
```diff
@@ -364,9 +364,18 @@ const authStateMachine = createMachine<AuthStateContext>(
...
+            {
+              target: [PATH_NAMES.ENTER_PASSWORD],
+              meta: { reversible: true },
+            },
+          ],
         },
         meta: {
           optionalPaths: [PATH_NAMES.ENTER_PASSWORD],
+          allowedEventsFromJourney: [
+            USER_JOURNEY_EVENTS.SIGN_IN_WITHOUT_PASSKEY,
```
BeckaL: Do we still need this optional path? Would be nice if this change allowed us to get rid of optional paths

BeckaL (commented):
This all looks OK to me, great job on small easy to review commits. One question here. Not approving as discussed - would be good to get another pair of eyes on this next week

src/app.ts
```diff
@@ -187,6 +188,7 @@ function registerRoutes(app: express.Application) {
     app.use(signInWithPasskeyRouter);
     app.use(accountExistsWithPasskeyRouter);
     app.use(cannotSignInPasskeyRouter);
+    app.use(journeyRouter);
```
ben-fernandes-sw, line 191: Should we feature flag this router?
(maybe put it behind an `||` of the passkey feature flags?)

src/components/journey/tests/journey-controller-integration.test.ts
```diff
@@ -66,4 +66,13 @@ describe("Integration::journey controller", () => {
       ]);
     });
   });
+
+  describe("validation failure", () => {
+    it("should redirect to session nextPath when page in route params does not match session", async () => {
+      await request(app)
+        .get("/journey/enter-password/SOME_OTHER_PAGE")
```
ben-fernandes-sw: Maybe make this test `/journey/some-other-page/A_VALID_EVENT` to make it clearer what we're testing here

src/components/journey/journey-controller.ts
```diff
@@ -1,4 +1,5 @@
 import type { Request, Response } from "express";
+import { authStateMachine } from "../common/state-machine/state-machine.js";
 import { getNextPathAndUpdateJourney } from "../common/state-machine/state-machine-executor.js";
 
 export async function journeyGet(req: Request, res: Response): Promise<void> {
```
ben-fernandes-sw, line 6: For each of the checks within this `journeyGet`, I think it might be nice to extract them into separate functions. So you can do something more like
```javascript
if (
  !routePageMatchesSessionPage(...)
  || !eventPermittedForPath(...)
  || !anotherCheck(...)
) {
  return res.redirect(previousPageFromSession)
}

// Happy path...
```

src/components/common/state-machine/state-machine.ts
```diff
@@ -364,9 +364,18 @@ const authStateMachine = createMachine<AuthStateContext>(
...
+              target: [PATH_NAMES.ENTER_PASSWORD],
+              meta: { reversible: true },
+            },
+          ],
         },
         meta: {
           optionalPaths: [PATH_NAMES.ENTER_PASSWORD],
+          permittedJourneyRouteEvents: [
+            USER_JOURNEY_EVENTS.SIGN_IN_WITHOUT_PASSKEY,
+          ],
```
ben-fernandes-sw, line 377: Just to get it in writing, I think this should exist within the `meta` object for each event, rather than attached to the state's `meta`.

Something like:
```javascript
[USER_JOURNEY_EVENTS.SIGN_IN_WITHOUT_PASSKEY]: [
  {
    target: [PATH_NAMES.ENTER_PASSWORD],
    meta: { reversible: true, permittedToUseJourney: true },
  },
],
```

But as we've discussed, happy to let this slide for now.

src/app.ts
ChristianSoftwire: It's currently behind the passkeyUsage feature flag and I think it should stay there for now (as we will only ever need to use `/journey` during passkey sign in, not passkey create)

ben-fernandes-sw (approved):

PR #3518, opened 2026-07-03
BAU: Uplift dompurify to 3.4.11 in journey-map

dbes-gds:
## What

Uplift dompurify to 3.4.11 in journey-map

Resolves multiple dependabot alerts

## How to review

1. Code Review
1. Start journey-map locally to test that it still works

## Checklist

- [x] Local running `./startup -L` still works.


sw-mattw (approved):
Tested running the journey map locally, seems to be working as expected

PR #3527, opened 2026-07-10
BAU: Refactor http to have default validate status

ChristianSoftwire:
## What

By default, axios treats any HTTP responses with status >= 200 && status < 300 as an error and throws an exception.
In http.ts, we explicitly add validateStatus where the status is between 200 and 400 (inclusive). This means, axios will not throw if the status code is 200 <= code <= 400. Instead, we handle these ourselves.

However, if you specify a validationStatuses array in the `getInternalRequestConfigWithSecurityHeaders` function, it overrides the default validateStatus specified in http.ts.
Replace this behaviour and instead add any extra status codes to the default. EG if you specify 401 as a validationStatus, axios will not throw when 200 <= code <= 400 and 401.

## How to review

1. Code review

### DIFF
--- a/src/utils/http.ts
+++ b/src/utils/http.ts
@@ -86,7 +86,13 @@ export function getInternalRequestConfigWithSecurityHeaders(
 
   if (options.validationStatuses) {
     config.validateStatus = function (status: number) {
-      return options.validationStatuses.includes(status);
+      const isWithinDefaultValidateStatus =
+        status >= HTTP_STATUS_CODES.OK &&
+        status <= HTTP_STATUS_CODES.BAD_REQUEST;
+      return (
+        isWithinDefaultValidateStatus ||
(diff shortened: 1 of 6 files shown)


src/utils/http.ts
```diff
@@ -86,7 +86,12 @@ export function getInternalRequestConfigWithSecurityHeaders(
 
   if (options.validationStatuses) {
     config.validateStatus = function (status: number) {
-      return options.validationStatuses.includes(status);
+      const defaultValidateStatus =
```
BeckaL: How about renaming this to `isWithinDefaultValidStatuses` or something like that?

BeckaL (dismissed):
Looks good to me, minor comment about renaming a variable

BeckaL (approved):

PR #3531, opened 2026-07-13
AUT-5548: Phased passkey rollout

ChristianSoftwire:
## What

Add a configurable env var that will be the threshold of journeys that will be prompted to register a passkey.
In /authorize, roll a dice and decide if this journey should include the passkey prompt. Save this value in the frontend session and use it in the register passkey prompting rules

## How to review

1. Code review


cloudformation/deploy/template.yaml
```diff
@@ -439,9 +464,11 @@ Mappings:
       SupportPasskeyUsage: "Yes"
       SupportPasskeyRegistration: "Yes"
       SupportNewInternationalSms: "No"
+      EnableDwpKbvContactFormChanges: "Yes"
       EnablePasskeyContactForm: "Yes"
       SupportSingleFactorAccountDeletion: "Yes"
       PasskeyPromptClientAllowList: "rPEUe0hRrHqf0i0es1gYjKxE5ceGN7VK"
+      PasskeyRolloutPercentage: 50
```
ChristianSoftwire: This is set to 50% for testing in authdev2. I will revert back to 100 once approved

src/config.ts
```diff
@@ -226,3 +230,7 @@ export function getPasskeyPromptClientAllowList(): string[] {
   }
   return [];
 }
+
+export function getPasskeyRolloutPercentage(): number {
+  return Number(process.env.PASSKEY_ROLLOUT_PERCENTAGE);
```
ben-fernandes-sw: Should handle the null `PASSKEY_ROLLOUT_PERCENTAGE` case here to avoid `Number` possibly throwing errors

src/utils/passkeys-helper.test.ts
```diff
@@ -112,9 +132,10 @@ describe("passkeys helper", () => {
         reauthenticate,
         rpClientId,
         backendIndicatesPasskeyPromptShouldBeSkipped,
+        isInPasskeyPhasedRollout,
         expected,
       }) => {
-        it(`should return ${expected} when browserSupportsWebAuthn=${browserSupportsWebAuthn}, hasActivePasskey=${hasActivePasskey}, hasSkippedPasskeyRegistration=${hasSkippedPasskeyRegistration}, supportPasskeyRegistration=${supportPasskeyRegistration}, reauthenticate=${reauthenticate}, rpClientId=${rpClientId}, backendIndicatesPasskeyPromptShouldBeSkipped=${backendIndicatesPasskeyPromptShouldBeSkipped}`, () => {
+        it(`should return ${expected} when browserSupportsWebAuthn=${browserSupportsWebAuthn}, hasActivePasskey=${hasActivePasskey}, hasSkippedPasskeyRegistration=${hasSkippedPasskeyRegistration}, supportPasskeyRegistration=${supportPasskeyRegistration}, reauthenticate=${reauthenticate}, rpClientId=${rpClientId}, backendIndicatesPasskeyPromptShouldBeSkipped=${backendIndicatesPasskeyPromptShouldBeSkipped}, isInPasskeyPhasedRollout=${isInPasskeyPhasedRollout}`, () => {
```
ben-fernandes-sw: This function name is getting a bit unwieldy. Could we change it to something a bit more like

Could we change the test case to look like this:
```javascript
testCase = {
  conditions: {
    browserSupportsWebAuthn: true,
    ....
  }
  expected: true
}
```

and the `it` to look like this:
```javascript
it(`should return ${expected} when conditions=${JSON.stringify(conditions)}`, () => {...})
```

ben-fernandes-sw (commented):
Looking good! Have left some small comments.

ben-fernandes-sw:
Tested 5 journeys in authdev2 and was prompted 2 times, with the rollout set to 50%

ben-fernandes-sw (approved):

PR #3532, opened 2026-07-14
Aut 5641: handle account interventions response from amc

BeckaL:
If amc return to the create passkey callback with an indication that the user has account interventions placed on them, we should redirect them to the relevant account interventions error screen.

We do this by bumping them straight to auth code, NOT going via sign in end. This means that they are not prompted for t&cs if relevant, or re-prompted to set up a passkey, but instead we let auth code handle the account interventions.

If the intervention is for re-prove identity, this will mean we successfully issue an auth code and hand back to orch, who handle this intervention.

## Related PRs

Adding this response to the stub: https://github.com/govuk-one-login/authentication-stubs/pull/776
Acceptance tests (to be merged after this one): https://github.com/govuk-one-login/authentication-acceptance-tests/pull/953


src/components/common/state-machine/state-machine.ts
```diff
@@ -637,6 +638,9 @@ const authStateMachine = createMachine<AuthStateContext>(
           [USER_JOURNEY_EVENTS.CREATE_PASSKEY_BACK]: [
```
ChristianSoftwire: Nitpick: would be nice to stick with the AMC_RETURNED_ convention for the other events

ChristianSoftwire (dismissed):
lgtm

ChristianSoftwire (approved):

PR #3548, opened 2026-07-15
AUT-5674: Bug fix - couldn't sign in with passkey after validation error

ben-fernandes-sw:
## What

When rendering the validation error screen of the `/cannot-sign-in-passkey` page, we weren't passing in the `authenticationOptions` to the `postValidationLocals` function.

This change generates the `authenticationOptions` as before within the GET controller, but then saves it in `req.session.user.cannotSignInPasskeyAuthOptions`. This can then be used later in the validation screen.

Refreshing the page will still load new options, as before.

## How to review

1. Code Review
1. Test in authdev3 (see repro steps on [the Jira ticket](https://govukverify.atlassian.net/browse/AUT-5674))

### DIFF
--- a/src/components/cannot-sign-in-passkey/tests/cannot-sign-in-passkey-integration.test.ts
+++ b/src/components/cannot-sign-in-passkey/tests/cannot-sign-in-passkey-integration.test.ts
@@ -37,6 +37,7 @@ describe("Integration:: cannot sign in passkey", () => {
               commonVariables.diPersistentSessionId;
 
             req.session.user = {
+              ...req.session.user,
               journey: getPermittedJourneyForPath(
                 PATH_NAMES.CANNOT_SIGN_IN_PASSKEY
               ),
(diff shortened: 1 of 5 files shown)


src/components/cannot-sign-in-passkey/tests/cannot-sign-in-passkey-integration.test.ts
```diff
@@ -37,6 +37,7 @@ describe("Integration:: cannot sign in passkey", () => {
               commonVariables.diPersistentSessionId;
 
             req.session.user = {
+              ...req.session.user,
```
BeckaL, line 40: Are the changes in this commit the tests related to your prod change in the first commit? If so can we squash the commits together so the commit is atomic?

src/components/templates/pages.ts
```diff
@@ -63,13 +63,15 @@ export const pages: Record<string, Page | PageVariant[]> = {
       name: "1FA journey",
       template: "cannot-sign-in-passkey/index.njk",
       options: {
+        authenticationOptions: "",
```
BeckaL, line 66: probs squash this in as well, part of the commit being in a working state

BeckaL (approved):

PR #3553, opened 2026-07-16
AUT-5044: suppress passkey setup prompt for users with account interventions call ticf

ChristianSoftwire:
## What

Before handing off to AMC during a passkey create journey, we want to do an AIS and TICF call and address any account interventions.

This PR covers doing those interventions checks before handing off to AMC. These are all handled in a similar way to auth_code (and we can see that in the state-machine). However, for reprove identity intervention, orch handles this, so we redirect to auth_code and orch will handle this intervention.

## How to review

1. Code review


src/components/auth-code/auth-code-routes.ts
```diff
@@ -15,7 +15,7 @@ router.get(
   validateSessionMiddleware,
   requiredSessionFieldsMiddleware,
   allowUserJourneyMiddleware,
-  accountInterventionsMiddleware(false, true, true),
+  accountInterventionsMiddleware(false, true, false, true),
```
BeckaL: I think there's a case here to add a refactoring to make the `handleX` options an object so it's clearer what's going on here

BeckaL (approved):

PR #3556, opened 2026-07-20
AUT-5643: Add reason to google analytics page view data

ChristianSoftwire:
## What

We need to add the "reason" field to google analytics page view data as part of the passkey initiative.
This reason field comes from any `authenticationErrors` we get from webauthn. 
We also weren't acting on errors from webauthn on the `cannot-sign-in-passkey` page, so this PR also covers adding that.

## How to review

1. Code review
2. Check 'reason' is sent when there is a webauthn error (like cancelling the popup) but it isn't sent elsewhere

### DIFF
--- a/src/components/common/constants.ts
+++ b/src/components/common/constants.ts
@@ -185,3 +185,14 @@ export const JOURNEY_TYPE = {
   SIGN_IN: "sign-in",
   CREATE_ACCOUNT: "create-account",
 };
+
+export const VALID_WEBAUTHN_ERRORS = [
+  "NotAllowedError",
+  "AbortError",
+  "SecurityError",
+  "InvalidStateError",
+  "ConstraintError",
+  "DataError",
+  "NotSupportedError",
+  "UnknownError",
+];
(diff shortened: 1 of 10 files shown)


src/components/common/constants.ts
```diff
@@ -185,3 +185,14 @@ export const JOURNEY_TYPE = {
   SIGN_IN: "sign-in",
   CREATE_ACCOUNT: "create-account",
 };
+
+export const VALID_WEBAUTHN_ERRORS = [
```
BeckaL, line 189: How did you get this list? Is there any way we can pull it directly from webauthn?

BeckaL (commented):
Looks sensible to me. Question about whether we can get errors from the webauthn library so our code less brittle

src/components/common/constants.ts
ChristianSoftwire: This is a list of the standard errors that can come from webauthn (https://www.corbado.com/blog/webauthn-errors#2-a-production-cheat-sheet) - doesn't seem to be a way to get this from the simplewebauthn/browser package. FWIW, these errors come from the browser, the library just wraps them

BeckaL (approved):

PR #3566, opened 2026-07-23
AUT-5686: Users with account interventions in passkeys go to ts and cs before intervention

ChristianSoftwire:
## What

Previously, users with account interventions on a passkey registration journey were correctly being checked for interventions, but were not being showed the T&Cs screen. 

This PR takes those users to the sign-in-end state and lets that handle showing the T&Cs screen.
This affects both the auth owned create-passkey page and the create-passkey-callback which handles any interventions from AMC

## How to review

1. Code review
2. Test an account with interventions and needing to update T&Cs gets taken to the T&Cs before getting an intervention


test/unit/middleware/account-interventions-middleware.test.ts
```diff
@@ -350,13 +350,46 @@ describe("accountInterventionsMiddleware", () => {
...
+      before(() => {
+        accountInterventionsWithBlockedTrue = accountInterventionsFakeHelper({
+          passwordResetRequired: false,
+          blocked: true,
+          temporarilySuspended: false,
+          reproveIdentity: false,
+        });
+      });
+
+      it("should call onRedirect before redirecting when property in options object", async () => {
```
ben-fernandes-sw, line 366: I'm a little worried about over-doing it in this test.
1. Do we actually care about what's happening within the `onRedirect` function? Can we not just pass a test function and check it was invoked once?
2. Do we need to check that the redirect happened?

ben-fernandes-sw (approved):

PR #3596, opened 2026-08-04
Aut 5528/update skip user profile

BeckaL:
Call update user profile when a user clicks skip in auth or amc, with an update type of passkey setup skipped.

This will ensure the field is set on the user profile, which will subsequently be used in the prompting rules to suppress prompts for users who have recently skipped adding a passkey.

Do not surface an error to the user if this call is not successful (better that they are prompted than encounter an error in their journeys).

Also some refactors to remove redundant types.

## How to test

Currently deployed to authdev2 alongside the backend change. 
Run through a journey where you're prompted to set up a passkey, and click skip on either page (the first auth passkey prompt page, or the amc stub page) and see that the timestamp is updated in the user profile table.

## Related PRs:

[Backend PR](https://github.com/govuk-one-login/authentication-api/pull/8644) - must be merged first - that added the ability to update the user profile in this way (and also removes the redundant types / api request fields seen here).


BeckaL:
Code coverage complaint here is erroneously reporting no coverage on lines where this is test coverage

src/components/create-passkey/create-passkey-controller.ts
```diff
@@ -31,10 +35,32 @@ export function createPasskeyPost(
...
+    email,
+    UpdateType.SKIP_ADDING_PASSKEY,
+    persistentSessionId,
+    req
+  );
+
+  if (!updateProfileResult.success) {
+    req.log.warn(
+      "Did not successfully manage to update profile with skip event, passkey registration prompt may not be suppressed"
+    );
```
ben-fernandes-sw, line 61: I wonder if there's any alarming we can set up for this

ben-fernandes-sw (approved):

src/components/create-passkey/create-passkey-controller.ts
BeckaL: Can do in splunk 👍

PR #3613, opened 2026-08-18
BAU: Passkey prompt logging

Q-GDS:
## What

The log lines in `shouldPromptToRegisterPasskey` sat after each early return, so they only ran when that check had passed. Lines still reached Splunk, but they recorded the checks a user got past rather than the one that stopped them. Working out why someone was not prompted meant noticing which line was missing and knowing the order the checks run in.

Most of the messages were also inaccurate. They were named after the condition rather than what was true when they ran, so `!isInPasskeyPhasedRollout=true` was logged when the user was in the rollout. 

Each guard now logs on the path where it stops the prompt and says why:

Passkey registration not enabled
User not in passkey phased rollout
Browser does not support WebAuthn, cannot register a passkey
User already has a passkey, or it cannot be determined
User has already skipped passkey registration
Backend indicated the passkey prompt should be skipped
User is reauthenticating, skipping passkey prompt
User has been on a password reset journey, skipping passkey prompt
Relying party is on the passkey prompt deny list
User has not signed in with a password and 2FA, skipping passkey prompt
Account intervention applied during passkey registration

A second commit adds a log to `isInPasskeyPhasedRollout`. It returns false both when the rollout percentage is unset and when the user falls outside it and the prompt logging says `User not in passkey phased rollout` for both. The new line tells the two apart.

## How to review

1. Code review

### DIFF
--- a/src/components/authorize/authorize-controller.ts
+++ b/src/components/authorize/authorize-controller.ts
@@ -203,7 +203,7 @@ export function authorizeGet(
       req.query[faceToFaceRpGoogleAnalyticsParamKey] as string
     );
 
-    req.session.user.isInPasskeyPhasedRollout = isInPasskeyPhasedRollout();
+    req.session.user.isInPasskeyPhasedRollout = isInPasskeyPhasedRollout(req);
 
     return res.redirect(
       appendQueryParamIfHasValue(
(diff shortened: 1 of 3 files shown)


BeckaL:
Happy to ignore sonar cloud

BeckaL (approved):

PR #3618, opened 2026-08-25
BAU: "/resend-code-account-creation" cleanup

ben-fernandes-sw:
## What

There was a lot of hangover code from the copying of `/resend-code` to `/resend-code-create-account`. This PR removes the unused `isResendCodeRequest` and where it's passed around.

**Where `isResendCodeRequest` comes from and why it's not needed**
1. `/resend-code-create-account?isResendCodeRequest=true` is set as a link in the `/check-your-phone` template
2. That query param is used by the `/resend-code-create-account` controller and sets `isResendCodeRequest` as a hidden field on the form
3. When _that_ form is submitted, `isResendCodeRequest` is never used

so there's no need for it to be passed down that chain.

## How to review

1. Code Review
6. Test doing an account creation journey and resending an SMS
7. Test completing a partially created account and resending an SMS


BeckaL (approved):

csophocleous (approved):
LGTM

